PR #45, opened 2022-11-28
Support ip blocks

adisos:


pkg/netpol/eval/check.go
```diff
@@ -230,6 +235,18 @@ func (pe *PolicyEngine) getPeer(p string) (k8s.Peer, error) {
 		}
 		return k8s.Peer{}, fmt.Errorf("could not find peer %s", p)
 	}
+	// check if p is an ip block (cidr)
+	if strings.Contains(p, "/") {
```
zivnevo, line 239: Seems like we can never get into this `if` because of the previous `if`.

pkg/netpol/eval/resources.go
```diff
@@ -174,3 +174,38 @@ func (pe *PolicyEngine) deleteNetworkPolicy(np *netv1.NetworkPolicy) error {
...
+			policyIPBlocksList := policy.GetReferencedIPBlocks()
+			ipbList = append(ipbList, policyIPBlocksList...)
+		}
+	}
+	newAll, _ := k8s.NewIPBlock("0.0.0.0/0", []string{})
+	disjointRes := k8s.DisjointIPBlocks(ipbList, []*k8s.IPBlock{newAll})
+	return disjointRes
+}
+
+// IsPeerIPType retunrs true for an input peer if it is an IP type
```
zivnevo, line 208:
```suggestion
// IsPeerIPType returns true for an input peer if it is an IP type
```
typo

pkg/netpol/eval/resources.go
```diff
@@ -174,3 +174,38 @@ func (pe *PolicyEngine) deleteNetworkPolicy(np *netv1.NetworkPolicy) error {
...
+	}
+	for podStr := range pe.podsMap {
+		if podPeer, err := pe.getPeer(podStr); err == nil {
+			res = append(res, &podPeer)
+		}
+	}
+	return res
+}
+
+// GetDisjointIPBlocks returns a slice of disjoint ip-blocks from all netpols reources
```
zivnevo, line 194:
```suggestion
// GetDisjointIPBlocks returns a slice of disjoint ip-blocks from all netpols resources
```
typo

zivnevo (changes requested):

zivnevo (dismissed):

pkg/netpol/eval/internal/k8s/ipBlock.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net"
 	"sort"
 	"strconv"
+	"strings"
 	// "inet.af/netaddr"
```
elevran, line 10: delete if unused.

pkg/netpol/eval/internal/k8s/ipBlock.go
```diff
@@ -23,15 +24,25 @@ type IPBlock struct {
 	ipRange CanonicalIntervalSet
 }
 
-// ToIPRanges returns a string if the ip ranges in the current IPBlock object
+// ToIPRanges returns a string of the ip ranges in the current IPBlock object
 func (b *IPBlock) ToIPRanges() string {
```
elevran, line 28: I wonder if there might be a struct and methods you can reuse from [netip](https://pkg.go.dev/net/netip) instead of roll your own. Not sure if they overlap fully or not (eg can any and all blocks be represented by a network.

pkg/netpol/eval/internal/k8s/ipBlock.go
```diff
@@ -6,6 +6,7 @@ import (
 	"net"
```
elevran, line 6: should the package for IPBlock be k8s? Seems unrelated.

pkg/netpol/eval/check.go
```diff
@@ -170,12 +176,12 @@ func (pe *PolicyEngine) allallowedXgressConnections(src, dst k8s.Peer, isIngress
 	// relevant policies: policies that capture dst if isIngress, else policies that capture src
 	var netpols []*k8s.NetworkPolicy
 	if isIngress {
-		if dst.PeerType == k8s.Iptype {
+		if dst.PeerType == k8s.IPBlockType {
```
elevran, line 179: may want to consider making `peer` an interface/struct since there seem to be variations on it

pkg/netpol/connlist/connlist.go
```diff
@@ -113,10 +113,18 @@ func FromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, err
 // getConnectionsList returns connections list from PolicyEngine object
 func getConnectionsList(pe *eval.PolicyEngine) ([]Peer2PeerConnection, error) {
 	res := []Peer2PeerConnection{}
```
elevran, line 115: - is `Peer2PeerConnection` used by other packages and should remain public?
- given the package name `connlist`, I wonder if `Peer2PeerConnection` is not too long of a name and whether we have the right abstraction level. Need to think about it, not required in this PR

pkg/netpol/eval/internal/k8s/peer.go
```diff
@@ -13,18 +13,31 @@
...
-	Iptype
+	IPBlockType
 )
 
 // Peer represents a k8s pod or an ip address
 type Peer struct {
-	PeerType  PeerType // PodType or Iptype
-	IP        string
+	PeerType  PeerType
+	IPBlock   *IPBlock // a set of intervals for ip addresses ranges
```
elevran, line 31: Pod and IPBlock are mutually exclusive - that is, only one or the other can be set, not both>
If so, seems like a good case for Peer interface, implemented for either mode

pkg/netpol/eval/internal/k8s/ipBlock.go
adisos: opened a separate issue #48

pkg/netpol/eval/internal/k8s/peer.go
adisos: opened a separate issue #49

zivnevo (approved):

pkg/netpol/eval/internal/k8s/ipBlock.go
adisos: Will consider in a separate issue to use ip-block representation from netip, and convert to ranges only for computation of disjoint ip-blocks.

PR #46, opened 2022-11-28
Add tests disjoint ip blocks

adisos:


pkg/netpol/connlist/connlist_test.go
```diff
@@ -0,0 +1,74 @@
...
+	outputFile := "out.txt"
+	for _, testName := range testNames {
+		path := filepath.Join(getTestsDir(), testName)
+		expectedOutputFile := filepath.Join(path, expectedOutputFileName)
+		res, err := FromDir(path)
+		if err != nil {
+			t.Fatalf("Test %v: TestConnList err: %v", testName, err)
+		}
+		writeRes(ConnectionsListToString(res), outputFile)
+		expectedHash, expectedErr := getFileHashValue(expectedOutputFile)
```
elevran, line 24: replace with `filesHaveSameContent(f1, f2)` or similar?
You can also do `stringsHaveSameContent(s1, s2)` by using `ioutil.ReadAll` from a file and avoid using `writeRes` in the previous line

elevran: Also, you may want to consider using a "golden file" package to compare and update test results against expected/golden files.
For example, [golden](https://pkg.go.dev/gotest.tools/v3/golden) or [autogold](https://github.com/hexops/autogold)

elevran (commented):
two suggestions for possible enhancements, nothing preventing a merge of the current PR

pkg/netpol/connlist/connlist_test.go
adisos: changed to comparison of strings, using `os.ReadFile`.

zivnevo (approved):

Issue #50, opened 2022-12-11
connlist required changes


adisos:
- [x] Peer2PeerConnection: change into an interface, and do not expose the internal fields.
Src/Dst: create another interface called Peer, with 3 methods: getName, getNamesmace, getIP, so it is clear for the consumer if it is a name/namespace combination or an ip address.
Ports: change to PortRange, with startPort and endPort. When they are  equal it represents a single port. When they are both equal to 0 it means that the port range is empty.

- [x] Library input: currently it is FromDir and FromK8sCluster. should expect a list of YAML manifests. (extracted from dir or cluster). Separate to have one function that extracts the YAML, and another function that calls the PolicyEngine, with the YAML inputs (slice of strings, where each string is a YAML document).
The function that should accept a list of YAML files should also be exported. (FromYAMLManifests). 
FromDir: add the option to add a custom walker. The consumer will provide a walker object (that can prevent scanning certain dirs for example).

Smaller comments: 
- [x] Add package documentation.
- [x] context.TODO() : consider replacing with context.WithTimeout().
- [x] ConnectionSet.GetProtocolsAndPortsMap() : return explicit type of port range instead of list of ports (which currently is always expected to have exactly two ports values). 
- [x] consistent slice/map initialization.
- [x] new line separator '\n' : should be tested on all systems. Can instead use a constant newline separator of a standard library.
- [x] Output printing : do not use %v for printing numbers, be more specific with %d. 
- [x] separator const var: use a more specific name, such as portRangeSeparator

Issue #49, opened 2022-12-01
change Peer type to an interface


adisos:
Pod and IPBlock are mutually exclusive - that is, only one or the other can be set, not both
change to Peer interface, implemented for either mode

PR #54, opened 2022-12-12
Peer as interface

adisos:
Handle issue #49 (change Peer type to an interface for package eval) + #50 (use eval.Peer interface instead of redefining a peer interface).

### DIFF
--- a/cmd/netpolicy/cmd/evaluate.go
+++ b/cmd/netpolicy/cmd/evaluate.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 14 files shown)


zivnevo:
I'm a bit confused. Can you please explain why we need 2 `Peer` interfaces? Why can't we just have the one in `eval`?

adisos:
> 

We need the Peer interface to be used internally at internal/k8s pkg and to be exposed externally by the eval pkg.
The interface for internal pkg is different from the interface of the eval pkg.
The eval.Peer interface should not have all methods of the k8s.Peer interface.
The internal interface is to handle issue #49 - convert Peer struct to Peer interface, and is used at netpol analysis.
The eval.Peer interface is to expose a Peer outside the eval pkg, such as connlist, and as the objects returned from eval.GetPeersList.

zivnevo (approved):
You might want to consider moving some of the code in `eval/check.go` to `internal/k8s/peer.go`, e.g., `isPodToItself()` and `isPeerNodeIP()`. Ideally, eval should work on a more abstract level.

Also, note that some of the files (e.g., `ipBlock.go` are missing the license header. Maybe open a separate issue for this.

adisos:
> You might want to consider moving some of the `k8s.Peer`-related code in `eval/check.go` to `internal/k8s/peer.go`, e.g., `isPodToItself()` and `isPeerNodeIP()`. Ideally, eval should work on a more abstract level.
> 
> Also, note that some of the files (e.g., `ipBlock.go`) are missing the license header. Maybe open a separate issue for this.

Added these comments to a separate issue.

PR #59, opened 2023-01-03
scan pkg - support workloads

adisos:

### DIFF
--- a/pkg/netpol/scan/scan.go
+++ b/pkg/netpol/scan/scan.go
@@ -2,6 +2,7 @@ package scan
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -10,6 +11,7 @@ import (
 
 	yamlv3 "gopkg.in/yaml.v3"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	netv1 "k8s.io/api/networking/v1"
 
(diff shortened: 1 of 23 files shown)


pkg/netpol/scan/scan.go
```diff
@@ -109,43 +134,112 @@ const (
...
-	res := make([]string, 0)
-	manifestFiles := searchDeploymentManifests(repoDir, walkFn)
-	for _, mfp := range manifestFiles {
-		filebuf, err := os.ReadFile(mfp)
-		if err != nil {
-			continue
-		}
-		res = append(res, splitByYamlDocuments(filebuf)...)
+// given a YAML doc and its resource kind, convert to a slice of K8sObject
+func mainfestToK8sObjects(yamlDoc, kind string) ([]K8sObject, error) {
```
zivnevo, line 194:
```suggestion
func manifestToK8sObjects(yamlDoc, kind string) ([]K8sObject, error) {
```

pkg/netpol/scan/scan.go
```diff
@@ -109,43 +134,112 @@ const (
...
+		// ignore empty cases
+		return false, ""
+	}
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	_, groupVersionKind, err := decode([]byte(yamlDoc), nil, nil)
+	if err != nil {
+		return false, ""
+	}
+	if !acceptedK8sTypes.MatchString(groupVersionKind.Kind) {
+		fmt.Printf("Skipping object with type: %s", groupVersionKind.Kind)
```
zivnevo, line 187:
```suggestion
		fmt.Printf("Skipping resource of kind: %s", groupVersionKind.Kind)
```

pkg/netpol/scan/scan.go
```diff
@@ -109,43 +134,112 @@ const (
...
+		return res, fmt.Errorf("unsupported kind: %s", kind)
 	}
-	return res
+	return res, nil
 }
 
+const yamlParseBufferSize = 200
+
 // return a list of paths for yaml files in the input dir path
 func searchDeploymentManifests(repoDir string, walkFn WalkFunction) []string {
```
zivnevo, line 244: Function name is misleading - it's not only deployments this function searches for.
Better call it `searchYamlFiles`.

pkg/netpol/scan/scan.go
```diff
@@ -164,69 +258,241 @@ func searchDeploymentManifests(repoDir string, walkFn WalkFunction) []string {
...
+	return nil
+}
+
+// given YAML of kind "PodList" , parse and convert to slice of K8sObject
+func parsePodList(objDataBuf []byte) []K8sObject {
+	if isKind, resList := parseListOfKind(objDataBuf, Pod); isKind {
+		return resList
+	}
+	return nil
+}
```
zivnevo, line 285: Why are pods and namespaces so special, they get a `parse*List` of their own? Is it because of the tests we have?
I believe all the kinds we care about can have their *List.
In general, `kubectl get` always returns an object of type `List`; It's API GET calls that return the `*List` kinds.
See [here](https://kubernetes.io/docs/reference/using-api/api-concepts/#collections)

zivnevo (changes requested):

pkg/netpol/scan/scan.go
adisos: Yes, it's because of the tests that we have, one test requires support in PodsList and NamespaceList.
I wrote in a TODO comment that we should add support for other `*List` kinds.

pkg/netpol/scan/scan_test.go
```diff
@@ -0,0 +1,135 @@
...
+    - env:
+      - name: PORT
+        value: "9555"
+      - name: DISABLE_STATS
+        value: "1"
+      imagePullPolicy: IfNotPresent
+      livenessProbe:
+        exec:
+          command:
+          - /bin/grpc_health_probe
```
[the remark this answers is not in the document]
adisos: done

tests/onlineboutique_workloads/replicaset.yaml
```diff
@@ -0,0 +1,1127 @@
+apiVersion: v1
+items:
+- apiVersion: apps/v1
+  kind: ReplicaSet
+  metadata:
+    annotations:
+      deployment.kubernetes.io/desired-replicas: "1"
+      deployment.kubernetes.io/max-replicas: "2"
+      deployment.kubernetes.io/revision: "1"
```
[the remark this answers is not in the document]
adisos: Added another test with other types of workloads as well.

pkg/netpol/scan/scan.go
adisos: done

pkg/netpol/scan/scan.go
adisos: done

pkg/netpol/scan/scan.go
adisos: fixed.

PR #61, opened 2023-01-03
support workloads: connlist output in level of workloads instead of pods

adisos:


zivnevo (approved):
Great work. I like how the `Peer` interface also captured workloads.

PR #63, opened 2023-01-10
 connlist updates: fix newline char + use context.WithTimeOut

adisos:

Remaining items from issue #50


zivnevo (dismissed):

cmd/netpolicy/cmd/evaluate.go
```diff
@@ -127,8 +128,13 @@ var evaluateCmd = &cobra.Command{
 		} else {
 			// get relevant resources from k8s live cluster
 			var err error
+			const ctxTimeoutSeconds = 3
+			ctx := context.Background()
+			ctx, cancel := context.WithTimeout(ctx, ctxTimeoutSeconds*time.Second)
```
elevran, line 133: `ctx` is being overridden.  Can refer to `context.Background` in the call when inializing the actual context

pkg/netpol/connlist/connlist.go
```diff
@@ -110,8 +112,12 @@ func FromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, err
 
 	// get all resources from k8s cluster
 
+	ctx := context.Background()
```
elevran, line 115: Reused var name in next line. No need to create it separately

pkg/netpol/connlist/connlist.go
```diff
@@ -188,7 +194,8 @@ func ConnectionsListToString(conns []Peer2PeerConnection) string {
 		connLines[i] = conns[i].String()
 	}
 	sort.Strings(connLines)
-	return strings.Join(connLines, "\n")
+	newlineChar := fmt.Sprintln("")
```
elevran, line 197: Not sure I understand the problem with using `\n`
That is a line break in all platforms but Windows

elevran (changes requested):

pkg/netpol/connlist/connlist.go
elevran: Println is better if you care about Windows as a runtime platform

PR #81, opened 2023-02-05
handle missing namespaces on creation of policyEngine from dir resources

adisos:


pkg/netpol/eval/resources.go
```diff
@@ -78,7 +78,26 @@ func NewPolicyEngineWithObjects(objects []scan.K8sObject) (*PolicyEngine, error)
...
+	err = pe.resolveMissingNamespaces()
+	return pe, err
+}
+
+func (pe *PolicyEngine) resolveMissingNamespaces() error {
+	for _, pod := range pe.podsMap {
+		ns := pod.Namespace
+		if _, ok := pe.namspacesMap[ns]; !ok {
+			// create a ns object and upsert to PolicyEngine
+			nsObj := &corev1.Namespace{}
```
elevran, line 90: Using an object initialization syntax here might be more succinct:
```go
nsObj := &corev1.Namespace {
   Name: ns,
   Labels: map[string]string{
         "kubernetes.io/metadata.name": ns,
   },
}
```

elevran (dismissed):
Please see NS creation suggestion

zivnevo (dismissed):

pkg/netpol/eval/resources.go
adisos: changed, thanks.

zivnevo (approved):

PR #107, opened 2023-03-22
new flag 'output' - supporting json format

shireenf-ibm:
first commit - add new flag `output`,  that takes `txt/json` values, currently supported with `k8snetpolicy list` cmd only


shireenf-ibm:
hi @adisos , do we want to add the `output` flag (indicates the required output-format) to work with `eval` too? (currently commited working with `list`)

pkg/netpol/connlist/connlist.go
```diff
@@ -204,17 +206,57 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
...
+	const (
+		txtFormat  = "txt"
+		jsonFormat = "json"
+	)
+	// default if output parameter is empty
+	if output == "" {
+		output = txtFormat
+	}
+
+	if output == jsonFormat {
```
elevran, line 221: may be better to validate format when getting command line parameters.

pkg/netpol/connlist/connlist.go
```diff
@@ -204,17 +206,57 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
 	return ca.getConnectionsList(pe)
 }
 
-// ConnectionsListToString returns a string of connections from list of Peer2PeerConnection objects
-func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection) string {
+// ConnectionsListToString returns a string of connections from list of Peer2PeerConnection objects in the required output format
+func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection, output string) (string, error) {
```
elevran, line 210: Pass a formatter (function or interface) instead of the format name as parameter?

pkg/netpol/connlist/connlist.go
```diff
@@ -204,17 +206,57 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
...
+	// supported output formats
```
elevran, line 211: This seems to better belong where the flag is defined.

cmd/netpolicy/cmd/list.go
```diff
@@ -66,4 +72,6 @@ func init() {
 	// output options
 	listCmd.Flags().StringVarP(&focusWorkload, "focusworkload", "",
 		focusWorkload, "Focus connections of specified workload name in the output")
+	// output format - default txt
+	listCmd.PersistentFlags().StringVarP(&output, "output", "o", "txt", "Required output format (txt, json) (default value: txt)")
```
elevran, line 76: I think it knows the default based on the passed in value of `txt` so no need to repeat in the option's explanation.

pkg/netpol/connlist/connlist.go
```diff
@@ -204,17 +206,57 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
...
+	const (
+		txtFormat  = "txt"
+		jsonFormat = "json"
+	)
+	// default if output parameter is empty
+	if output == "" {
```
elevran, line 217: the default is already set in the flag's definition.

pkg/netpol/connlist/connlist.go
```diff
@@ -204,17 +206,57 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
...
-		connLines[i] = conns[i].String()
+		connLines[i] = conns[i].TxtFormat()
 	}
 	sort.Strings(connLines)
 	newlineChar := fmt.Sprintln("")
 	return strings.Join(connLines, newlineChar)
 }
 
+// ConnectionsListTojson returns a json string form of connections from list of Peer2PeerConnection objects
+func (ca *ConnlistAnalyzer) ConnectionsListTojson(conns []Peer2PeerConnection) (string, error) {
```
elevran, line 242: Go conventions prefer `JSON` in the function's name.

pkg/netpol/connlist/connlist.go
```diff
@@ -229,6 +271,8 @@ type Peer2PeerConnection interface {
 	ProtocolsAndPorts() map[v1.Protocol][]eval.PortRange
 	// String returns a string representation of the connection object
 	String() string
+	TxtFormat() string
```
elevran, line 274: Consider using an external formatter/function accepting a connection and returning a string (kind of like the way Marshallers work - you may even be able to use the default JSON marshaller for this...)

pkg/netpol/connlist/connlist.go
```diff
@@ -260,24 +304,45 @@ func (c *connection) ProtocolsAndPorts() map[v1.Protocol][]eval.PortRange {
 	return c.protocolsAndPorts
 }
 
-// return a string representation for a connection object
+// return a txt representation for a connection object
+func (c *connection) TxtFormat() string {
+	connStr := c.String()
```
elevran, line 309: Is there a reason not to use `c.String()` as the text format?

pkg/netpol/connlist/connlist.go

pkg/netpol/connlist/connlist.go
shireenf-ibm: done - created a new interface - `ConnsFormatte`r , implemented by `TxtFormatter` and `JSONFormatter` and writes the output based on the output arg

pkg/netpol/connlist/connlist.go
shireenf-ibm: moved these to the new file too, tried to avoid code duplication by using the "json" struct for both formats

pkg/netpol/connlist/connlist.go
shireenf-ibm: actually `String()` only converted the protocols and ports fields of the connection to string ,replaced by `getProtocolsAndPortsStr()`

pkg/netpol/connlist/connlist.go
```diff
@@ -204,15 +203,9 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
...
-func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection) string {
-	connLines := make([]string, len(conns))
-	for i := range conns {
-		connLines[i] = conns[i].String()
-	}
-	sort.Strings(connLines)
-	newlineChar := fmt.Sprintln("")
-	return strings.Join(connLines, newlineChar)
+// ConnectionsListToString returns a string of connections from list of Peer2PeerConnection objects in the required output format
+func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection, outputFormat ConnsFormatter) (string, error) {
```
adisos, line 207: `outputFormat` should be a member of `ConnlistAnalyzer` instead of argument to this function.
Should use the type of `ConnlistAnalyzerOption` to add a new option as function `WithOutputFormat`.

adisos (changes requested):
looks good, added some comment

pkg/netpol/connlist/connlist.go
```diff
@@ -76,6 +77,12 @@ func WithFocusWorkload(workload string) ConnlistAnalyzerOption {
 	}
 }
 
+func WithOutputFormat(outputFormat ConnsFormatter) ConnlistAnalyzerOption {
+	return func(p *ConnlistAnalyzer) {
+		p.outputFormat = outputFormat
+	}
+}
+
```
adisos, line 85: since this is the interface to be used also as a library from external users, I prefer that the `outputFormat` type here is `string` and converted to `ConnsFormatter` internally. Also `ConnsFormatter` can be internal to this package and not exported.

pkg/netpol/connlist/connlist_test.go
```diff
@@ -38,8 +38,8 @@ func TestConnList(t *testing.T) {
 		if err != nil {
 			t.Fatalf("Test %s: TestConnList FromDir err: %v", testName, err)
 		}
-		actualOutput := analyzer.ConnectionsListToString(res)
-
+		analyzer.outputFormat = TxtFormatter{}
+		actualOutput, _ := analyzer.ConnectionsListToString(res)
```
adisos, line 42: why do we need this change?

cmd/netpolicy/cmd/list.go
```diff
@@ -42,7 +45,12 @@ defined`,
 		var conns []connlist.Peer2PeerConnection
 		var err error
 
-		analyzer := connlist.NewConnlistAnalyzer(connlist.WithFocusWorkload(focusWorkload))
+		outputFormat, err := validateAndGetOutputFormat()
```
adisos, line 48: consider using a flag validation function to the command object.
Take a look at `cmd/evaluate.go`.

pkg/netpol/connlist/connlist.go

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: removed

cmd/netpolicy/cmd/list.go
```diff
@@ -66,4 +86,18 @@ func init() {
...
+// validate the value of output arg
+func validateOutputArg() error {
+	// possible values of output arg
+	validFormats := []string{defaultFormat, "json"}
+	for _, formatName := range validFormats {
+		if output == formatName {
+			return nil
+		}
+	}
+	return errors.New(output + " output format is not supported.")
```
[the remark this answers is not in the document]
adisos: should not use hard-coded `"json"` string value

pkg/netpol/connlist/connlist.go
```diff
@@ -204,15 +211,18 @@ func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clients
 	return ca.getConnectionsList(pe)
 }
 
-// ConnectionsListToString returns a string of connections from list of Peer2PeerConnection objects
-func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection) string {
-	connLines := make([]string, len(conns))
-	for i := range conns {
-		connLines[i] = conns[i].String()
+// ConnectionsListToString returns a string of connections from list of Peer2PeerConnection objects in the required output format
+func (ca *ConnlistAnalyzer) ConnectionsListToString(conns []Peer2PeerConnection) (string, error) {
```
adisos, line 215: This change has an impact on the exported interface of `ConnlistAnalyzer`, so should require a new version release and adjustments.

adisos: Maybe `getFormatter` should return error if the output format is not valid

adisos (changes requested):
Added a few more comments.

pkg/netpol/connlist/conns_formatter.go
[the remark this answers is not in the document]
adisos: And also with explicit `txt` and explicit invalid output format.

cmd/netpolicy/cmd/list.go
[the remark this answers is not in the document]
shireenf-ibm: moved this func to the `connlist.go` file, so it is called from the `getFormatter func `
however, in order to still validate the output arg here (in `PersistentPreRunE`), I made it exported.
if it is undesirable to make it exported, we may remove the  `PersistentPreRunE` here and keep the "output" validation for the "ConnectionsListToString" step

adisos (changes requested):

cmd/netpolicy/cmd/command_test.go
adisos, line 1: add also test of `illegal list cmd with invalid output format`
Expected output should contain the string `"output format is not supported"`

cmd/netpolicy/cmd/list.go
```diff
@@ -37,12 +40,23 @@ defined`,
   # Get list of allowed connections from live k8s cluster
   k8snetpolicy list -k ./kube/config`,
 
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error { // Call parent pre-run
```
adisos, line 43:
```suggestion
	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
```

adisos (approved):

Issue #65, opened 2023-01-17
add workflows


adisos:
- [x] dependabot
- [x] codeql-analysis
- [x] make release
- [x] tests
- [x] consider linter settings

PR #114, opened 2023-04-03
Tests workflow

shireenf-ibm:
#65


shireenf-ibm:
hi @adisos ,
I found a way to skip running the `TestGeneralPerformance` in `eval_test.go` on Github. (succeeded to skip as I see in the workflow 
log).

however, after first commit : we had another failing test which is `test_illegal_eval_no_args` in command_test.go
it fails since this test tries to run the cmd `eval` without any args, neither the -s/-d/-p of eval nor the --dirpath/ -k.
if --dirpath is missing , root.go tries to load the `kubeconfig` file, and fails since we don't have it/ any live cluster on Github server

so, in order to avoid this test failure , we may :
1. add --dirpath to the test
2. change the order of validating eval flags in the `PersistentPreRunE` of `evaluate.go` - first check eval flags, then call it parent `PersistentPreRunE` 

committed the second one for now and it works

cmd/netpolicy/cmd/evaluate.go
```diff
@@ -84,6 +78,13 @@ var evaluateCmd = &cobra.Command{
 			return errors.New("destination port name or value is required")
 		}
 
+		// Call parent pre-run
+		if rootCmd.PersistentPreRunE != nil {
+			if err := rootCmd.PersistentPreRunE(cmd, args); err != nil {
+				return err
+			}
+		}
+
```
adisos, line 87: should we also change the order in `list` cmd?

adisos (approved):

PR #115, opened 2023-04-04
dot output

shireenf-ibm:


pkg/netpol/connlist/conns_formatter.go
```diff
@@ -67,3 +69,56 @@ func (j jsonFormatter) writeOutput(conns []Peer2PeerConnection) (string, error)
...
+	}
+	var graphLines string
+	for _, peerLine := range peerLines {
+		graphLines += peerLine
+	}
+	for _, edgeLine := range edgeLines {
+		graphLines += edgeLine
+	}
+	res := "digraph {" + d.sortGraphLines(graphLines) + "\n}\n"
+	return res, nil
```
adisos, line 123: Should sort separately edge lines and peer lines , so that in the output peer lines are placed before edge lines.
Also, avoid having the current flow of strings concatenation, then splitting, then concatenating again.

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -67,3 +69,56 @@ func (j jsonFormatter) writeOutput(conns []Peer2PeerConnection) (string, error)
...
+	peerName := peer.String()
+	return fmt.Sprintf("\t%q [label=%q color=%q fontcolor=%q]\n", peerName, peerName, peerColor, peerColor)
+}
+
+// sorts the final dot graph lines
+func (d dotFormatter) sortGraphLines(graphLines string) string {
+	var graphLinesSorted sort.StringSlice = strings.Split(graphLines, "\n")
+	graphLinesSorted.Sort()
+	return strings.Join(graphLinesSorted, "\n")
+}
```
adisos, line 99: why do these functions take `dotFormatter` as a receiver?

tests/onlineboutique_workloads/connlist_output.dot
adisos, line 1: where are all the peer lines?

shireenf-ibm: lines 3, 15 are examples of peer-lines (mixed sort)
I used that way of sort since we can not sort map keys in golang (couldn't find a convenient way to do that)
however, we can sort the list members.
I'll think of a way to write the peersLines also in a list , then sort separately

pkg/netpol/connlist/conns_formatter.go
shireenf-ibm: I wanted to tell these are "internal" for `dotFormatter` use only - (thinking of it as a class) - my fault ':)

tests/onlineboutique_workloads/connlist_output.dot
adisos: yes, you can create another list from the keys of the map and then sort it.

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -67,3 +69,57 @@ func (j jsonFormatter) writeOutput(conns []Peer2PeerConnection) (string, error)
...
+	sort.Strings(edgeLines)
+
+	res := "digraph {\n"
+	for _, peerLine := range peerLines {
+		res += peerLine
+	}
+	for _, edgeLine := range edgeLines {
+		res += edgeLine
+	}
+	res += "}\n"
```
adisos, line 123: be consistent with `txtFormatter.writeOutput()` : use `strings.Join` instead of concatenation with `+` where possible, and use `fmt.Sprintln("")` to get newline char (separate to a function to avoid duplication).

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -67,3 +69,57 @@ func (j jsonFormatter) writeOutput(conns []Peer2PeerConnection) (string, error)
...
+	for index := range conns {
+		connLine := formSingleConn(conns[index])
+		edgeLines[index] = getEdgeLine(connLine)
+		if _, ok := peersVisited[connLine.Src]; !ok {
+			peersVisited[connLine.Src] = visited
+			peerLines = append(peerLines, getPeerLine(conns[index].Src()))
+		}
+		if _, ok := peersVisited[connLine.Dst]; !ok {
+			peersVisited[connLine.Dst] = visited
+			peerLines = append(peerLines, getPeerLine(conns[index].Dst()))
```
adisos, line 109:
```suggestion
	peerLines := make([]string, 0) // list of peers lines
	for index := range conns {
		connLine := formSingleConn(conns[index])
		edgeLines[index] = getEdgeLine(connLine)
		if _, ok := peersVisited[connLine.Src]; !ok {
			peersVisited[connLine.Src] = struct{}{}
			peerLines = append(peerLines, getPeerLine(conns[index].Src()))
		}
		if _, ok := peersVisited[connLine.Dst]; !ok {
			peersVisited[connLine.Dst] = struct{}{}
			peerLines = append(peerLines, getPeerLine(conns[index].Dst()))
```

adisos (changes requested):

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -5,8 +5,14 @@ import (
...
 	"sort"
 	"strings"
+
+	"github.com/np-guard/netpol-analyzer/pkg/netpol/eval"
 )
 
+func getNewLineChar() string {
+	return fmt.Sprintln("")
+}
+
```
adisos, line 15:
```suggestion

const (
	dotHeader  = "digraph {"
	dotClosing = "}"
)

func getNewLineChar() string {
	return fmt.Sprintln("")
}

```

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -67,3 +72,51 @@ func (j jsonFormatter) writeOutput(conns []Peer2PeerConnection) (string, error)
...
+		}
+	}
+	// sort graph lines
+	sort.Strings(peerLines)
+	sort.Strings(edgeLines)
+	dotHeader := "digraph {" + getNewLineChar()
+	dotClosing := getNewLineChar() + "}"
+	res := dotHeader + strings.Join(peerLines, getNewLineChar()) + getNewLineChar() +
+		strings.Join(edgeLines, getNewLineChar()) + dotClosing
+	return res, nil
```
adisos, line 121:
```suggestion
	// collect all lines by order
	allLines := []string{dotHeader}
	allLines = append(allLines, peerLines...)
	allLines = append(allLines, edgeLines...)
	allLines = append(allLines, dotClosing)
	return strings.Join(allLines, getNewLineChar()), nil
```

adisos (approved):

PR #133, opened 2023-05-07
Add openshift resources to the serializer scheme

shireenf-ibm:


zivnevo:
Latest CI failures seem to be related to using Go 1.18.
@shireenf-ibm , I suggest upgrading to Go 1.19 (in go.mod and in all workflows).
StackRox are already consuming other libraries that use Go 1.19.

.github/workflows/go-build.yml
```diff
@@ -18,13 +18,9 @@ jobs:
...
     - name: Build
-      env:
-        CGO_ENABLED: 0
-        GOOS: linux      
-      run: go build --tags static_all -v -o ./bin/netpolicy cmd/netpolicy/main.go
-    
+      run: make build
     - name: Test
-      run: go test -v ./...
+      run: make test
```
adisos, line 26: why do we need these changes?

shireenf-ibm: they are not necessary (I just was trying to figure if the build failure was connected to the commands or not).
however I think we better use these , this way we avoid duplication by writing the build and test commands only once (in the make file) and use them for both local and on server runs.

This way (using make) is used in cluster-topology-analyzer's workflow too

pkg/netpol/scan/scan.go
```diff
@@ -16,23 +16,33 @@ import (
...
 )
 
 type ResourcesScanner struct {
-	logger      logger.Logger
-	stopOnError bool
-	walkFn      WalkFunction
+	logger          logger.Logger
+	stopOnError     bool
+	walkFn          WalkFunction
+	resourceDecoder runtime.Decoder
```
adisos, line 33: why is `resourceDecoder` a member of `ResourcesScanner`?

adisos (approved):

PR #139, opened 2023-05-15
improving Lint options

shireenf-ibm:
related issue : #65
Describing changes in each commit of this PR:
1. first commit: syncing the linter's options with the options used in cluster-topology-analyzer.
2. second commit: resolving `misspell` lint errors ( 2 typo)
3. third commit : resolving `gocritic` lint errors , included following fixes:
     -  naming returned values for functions that return multiple values which are of same type. 
     `golang `  suggests naming the results.
     - eliminating if-elseif-else chains (it is a lint error - so tried to replace with `switch`/add` return` to eliminate the else when possible)
     - new octal literal style (e.g. write mode `0o600` instead of ` 0600`)
     - adding explanation of why we use `nolint` in the cases it is used (to skip raising lint error)
     
4. forth commit:  we had only one `funlen` lint error , I think it is better not to break the `func` into smaller ones, so skipped it using `nolint`
5. fifth commit : resolving `prealloc` lint errors , ( declaring lists using `make` instead of var []<type>)
6. sixth commit: the only lint error remained is `gochecknoinits`, in order to resolve it , i.e. to eliminate the use of `init` func for each command , we should add other `func`s and not use the commands as `global var` in the file.
Since this means refactoring the `cmd` files , I prefer to do it in a new PR (added a new issue : #141)
and skipped it here using `nolint`


adisos (approved):
looks good!

Issue #143, opened 2023-05-22
simplify connlist output: remove trivial connectivity lines when possible 


adisos:
Connections from pod to itself can be removed from the report. 
Connection from deployment to itself is considered trivial only if the deployment has one replica.

adisos:
To simplify output, decided to avoid having in the output connections from workload to itself, regardless of whether it is trivial or not.

PR #148, opened 2023-05-22
Remove_trivial_connectivity_lines

shireenf-ibm:
#143


pkg/netpol/eval/check.go
```diff
@@ -76,51 +76,59 @@ func (pe *PolicyEngine) convertWorkloadPeerToPodPeer(peer Peer) (*k8s.PodPeer, e
 
 // for connectivity considerations, when requesting allowed connections between 2 workload peers which are the same,
 // looking for 2 different pod instances, if exist (to avoid the trivial case of connectivity from pod to itself)
-func (pe *PolicyEngine) changePodPeerToAnotherPodObject(peer *k8s.PodPeer) {
+// returns true if different pod instances found
```
adisos, line 79:
```suggestion
// if different pod instance was found - returns true and updates the input peer with another pod instance
```

pkg/netpol/eval/check.go
```diff
@@ -76,51 +76,59 @@ func (pe *PolicyEngine) convertWorkloadPeerToPodPeer(peer Peer) (*k8s.PodPeer, e
...
 // expecting that srcPeer and dstPeer are in level of workloads (WorkloadPeer)
-func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, error) {
+func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, bool, error) {
+	isTrivial := false
 	if srcPeer.IsPeerIPType() && !dstPeer.IsPeerIPType() {
 		// assuming dstPeer is WorkloadPeer, should be converted to k8s.Peer
 		dstPodPeer, err := pe.convertWorkloadPeerToPodPeer(dstPeer)
 		if err != nil {
-			return nil, err
+			return nil, isTrivial, err
```
adisos, line 101:
```suggestion
			return nil, false, err
```

pkg/netpol/eval/check.go
```diff
@@ -76,51 +76,59 @@ func (pe *PolicyEngine) convertWorkloadPeerToPodPeer(peer Peer) (*k8s.PodPeer, e
...
+	return false
 }
 
 // AllAllowedConnectionsBetweenWorkloadPeers returns the allowed connections from srcPeer to dstPeer,
+// and a bool flag indicating if the allowed connections are trivial
+// i.e. Connection from deployment to itself when the deployment has only one replica.
 // expecting that srcPeer and dstPeer are in level of workloads (WorkloadPeer)
-func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, error) {
+func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, bool, error) {
+	isTrivial := false
```
adisos, line 96:
```suggestion
```
move the definition of is`Trivial` to be before the call to `changePodPeerToAnotherPodObject` , where it is relevant

pkg/netpol/eval/check.go
```diff
@@ -76,51 +76,59 @@ func (pe *PolicyEngine) convertWorkloadPeerToPodPeer(peer Peer) (*k8s.PodPeer, e
...
 		if srcPeer.String() == dstPeer.String() {
-			pe.changePodPeerToAnotherPodObject(dstPodPeer)
+			isTrivial = !pe.changePodPeerToAnotherPodObject(dstPodPeer)
 		}
-		return pe.allAllowedConnectionsBetweenPeers(srcPodPeer, dstPodPeer)
+		conn, err := pe.allAllowedConnectionsBetweenPeers(srcPodPeer, dstPodPeer)
+		return conn, isTrivial, err
 	}
-	return nil, fmt.Errorf("cannot have both srcPeer and dstPeer of IP types: src: %s, dst: %s", srcPeer.String(), dstPeer.String())
+	return nil, isTrivial, fmt.Errorf("cannot have both srcPeer and dstPeer of IP types: src: %s, dst: %s", srcPeer.String(), dstPeer.String())
```
adisos, line 131:
```suggestion
	return nil, false, fmt.Errorf("cannot have both srcPeer and dstPeer of IP types: src: %s, dst: %s", srcPeer.String(), dstPeer.String())
```

pkg/netpol/eval/check.go
```diff
@@ -76,51 +76,59 @@ func (pe *PolicyEngine) convertWorkloadPeerToPodPeer(peer Peer) (*k8s.PodPeer, e
...
 	}
+	return false
 }
 
 // AllAllowedConnectionsBetweenWorkloadPeers returns the allowed connections from srcPeer to dstPeer,
+// and a bool flag indicating if the allowed connections are trivial
+// i.e. Connection from deployment to itself when the deployment has only one replica.
 // expecting that srcPeer and dstPeer are in level of workloads (WorkloadPeer)
-func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, error) {
+func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer Peer) (Connection, bool, error) {
```
adisos, line 95: does it handle identifying trivial connections if the input peers are pods and not workloads? do we have such tests?

adisos (changes requested):
hi @shireenf-ibm , added a few comments.
The main question is - do we handle both the case of input workload resources, and the case of input pod resources?  please check and see if we have tests for both scenarios.
Also, I tried to check `tests\onlineboutique_with_replicas` and it seems this test contains both pods and deployments resources -- I think it should be split into the two cases separately..

shireenf-ibm:
> Also, I tried to check `tests\onlineboutique_with_replicas` and it seems this test contains both pods and deployments resources -- I think it should be split into the two cases separately..

we already had the `tests/onlineboutique_with_replicas_and_variants` which includes only pods resources.
so I removed `pods.yaml` from `tests\onlineboutique_with_replicas`, so it will include deployment resources only .
and added both to the `connlist_test.go ` so we can view their connections list.

for `tests\onlineboutique_with_replicas` , since it doesn't include pods resources any more, I had to remove its `tests/onlineboutique_with_replicas/connections_map_with_replicas_output.txt` and eliminate the relevant test in `eval_test.go`, as it creats an empty `podsMap` now.

pkg/netpol/eval/check.go
shireenf-ibm: seems to me yes it is, 
does the `podlist ` of `tests/ipblockstest` meets the requirement?

PR #152, opened 2023-06-01
remove connections from workload to itself regardless existence of replicas

shireenf-ibm:
#143


shireenf-ibm:
I chose to open a new PR, in order to 1. revert the changes to `eval_tests` of `onlineboutique_with_replicas` 
2. apply the new fix (no sel-loop connections regardless existence of replicas)

pkg/netpol/connlist/connlist.go
```diff
@@ -380,7 +380,12 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
 				ca.errors = append(ca.errors, newResourceEvaluationError(err))
 				return nil, err
 			}
-			// skip empty connections
+			// skip nil connenctions or empty connections
+			// nil connections are returned in case of self-looped connection,
+			// i.e. a connection from workload to itself (regardless existence of replicas)
+			if allowedConnections == nil {
+				continue
+			}
```
adisos, line 388: I think we can avoid calling `AllAllowedConnectionsBetweenWorkloadPeers` when `srcPeer` and `dstPeer` are the same workload, instead of having to check a special return value.

pkg/netpol/eval/check.go
```diff
@@ -115,10 +103,9 @@ func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPe
 		if err != nil {
 			return nil, err
 		}
-		// if src and dst are the same workload peer, their conversion to pods should be of different pods
-		// (if owner has more than 1 instances)
+		// if src and dst are the same workload peer, it is a self-loop connection we want to skip it
 		if srcPeer.String() == dstPeer.String() {
-			pe.changePodPeerToAnotherPodObject(dstPodPeer)
+			return nil, nil
```
adisos, line 108: I prefer to maintain here the previous returned value (based on `changePodPeerToAnotherPodObject`) , so that package `eval` does return the most accurate connectivity result, and only package `connlist` chooses to avoid evaluating allowed connections between two same workloads.

adisos (changes requested):

adisos:
Also, update documentation (both README and package documentation)  and example output in the README .

pkg/netpol/connlist/connlist.go
```diff
@@ -1,5 +1,7 @@
 // The connlist package of netpol-analyzer allows producing a k8s connectivity report based on network policies.
-// It lists the set of allowed connections between each pair of peers (k8s workloads or ip-blocks).
+// It lists the set of allowed connections between each pair of different peers (k8s workloads or ip-blocks).
+// Self looped connections such as trivial connections from a peer to itself,
+// or connections between different replicas of same workload will be excluded.
```
adisos, line 4:
```suggestion
// Connections between workload to itself are excluded from the output.
```

README.md
```diff
@@ -42,7 +42,9 @@ Global Flags:
 
 ### List command
 ```
-Lists all allowed connections based on the workloads and network policies defined
+Lists all allowed connections based on the workloads and network policies defined.
+Excluding self looped connections such as trivial connections from a peer to itself,
+or connections between different replicas of same workload.
```
adisos, line 47:
```suggestion
Connections between workload to itself are excluded from the output.
```

adisos (changes requested):

pkg/netpol/connlist/connlist.go
```diff
@@ -375,6 +376,11 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
 			if !ca.includePairOfWorkloads(srcPeer, dstPeer) {
 				continue
 			}
+			// skip self-looped connection,
+			// i.e. a connection from workload to itself (regardless existence of replicas)
+			if srcPeer.String() == dstPeer.String() {
+				continue
+			}
```
adisos, line 383: maybe this should also be checked within `includePairOfWorkloads()`

shireenf-ibm: right , moved it

adisos (approved):

Issue #151, opened 2023-06-01
error handling updates 


adisos:
change to warning instead of error the case where there are no network policies

PR #153, opened 2023-06-01
converting the noK8sNetworkPolicyResourcesFound to warning

shireenf-ibm:
#151


adisos (changes requested):
Please also add a test to check this warning is returned in such a case.

shireenf-ibm:
> Please also add a test to check this warning is returned in such a case.

we have a test capturing it. Added requirements to check it is warning (not fatal/severe error)

adisos (approved):

PR #158, opened 2023-06-12
initial support of k8s service

shireenf-ibm:
- in tests , service objects are counted (added to the len of parsed k8s objects )

### DIFF
--- a/pkg/netpol/eval/eval_test.go
+++ b/pkg/netpol/eval/eval_test.go
@@ -944,13 +944,12 @@ type TestEntry struct {
 	nsList      []*v1.Namespace
 	podsList    []*v1.Pod
 	policies    []*netv1.NetworkPolicy
-	services    []*v1.Service
 }
 
 func initTest(test *TestEntry, t *testing.T) (*PolicyEngine, error) {
 	pe := NewPolicyEngine()
-	if len(test.nsList) > 0 || len(test.podsList) > 0 || len(test.policies) > 0 || len(test.services) > 0 {
-		err := pe.SetResources(test.policies, test.podsList, test.nsList, test.services)
+	if len(test.nsList) > 0 || len(test.podsList) > 0 || len(test.policies) > 0 {
+		err := pe.SetResources(test.policies, test.podsList, test.nsList)
 		if err != nil {
 			t.Fatalf("error init test: %v", err)
 			return nil, err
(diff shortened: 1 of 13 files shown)


pkg/netpol/eval/eval_test.go
adisos, line 1: The tests in this file are not involving service objects, so I think we can avoid these changes.

pkg/netpol/eval/internal/k8s/service.go
```diff
@@ -0,0 +1,40 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package k8s
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Service encapsulates k8s service fields that are relevant for evaluating network policies
```
adisos, line 20:
```suggestion
// Service encapsulates k8s service fields that are relevant for connectivity analysis
```

pkg/netpol/eval/resources.go
```diff
@@ -110,7 +113,7 @@ func (pe *PolicyEngine) resolveMissingNamespaces() error {
 // Deprecated: this function simply calls UpsertObject on the PolicyEngine.
 // Calling the UpsertObject should be preferred in new code.
 func (pe *PolicyEngine) SetResources(policies []*netv1.NetworkPolicy, pods []*corev1.Pod,
-	namespaces []*corev1.Namespace) error {
+	namespaces []*corev1.Namespace, services []*corev1.Service) error {
```
adisos, line 116: this function is deprecated and only used in eval_test so I think we can currently avoid this change.

pkg/netpol/eval/resources.go
```diff
@@ -126,6 +129,11 @@ func (pe *PolicyEngine) SetResources(policies []*netv1.NetworkPolicy, pods []*co
 			return err
 		}
 	}
+	for i := range services {
+		if err := pe.upsertService(services[i]); err != nil {
+			return err
+		}
+	}
```
adisos, line 136: same here

pkg/netpol/eval/resources.go
```diff
@@ -234,6 +248,17 @@ func (pe *PolicyEngine) upsertNetworkPolicy(np *netv1.NetworkPolicy) error {
...
 }
 
+func (pe *PolicyEngine) upsertService(svc *corev1.Service) error {
+	svcObj, err := k8s.ServiceFromCoreObject(svc)
+	if err != nil {
+		return err
+	}
+	svcStr := types.NamespacedName{Namespace: svcObj.Namespace, Name: svcObj.Name}.String()
+	pe.servicesMap[svcStr] = svcObj
+	// todo : should clear cache on services changes?
```
adisos, line 258:
```suggestion
```
no, it is related to `eval` results cached per pods with same owner

adisos (changes requested):
hi @shireenf-ibm , looks good, added a few comments.
In the next PR let's add the mapping from service to its pods.

adisos (approved):

Issue #112, opened 2023-04-03
add support for Route and Ingress resources


adisos:
`connlist` command should consider connectivity inferred from `Route` and `Ingress` resources as well, combined with network policy connectivity analysis.
For example, consider the Route below:
```
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  name: hello-world
  namespace: helloworld
spec:
  port:
    targetPort: 8000
  to:
    kind: Service
    name: hello-world
    weight: 100
```

The report of `connlist` may contain a line as follows:
`{ingress-controller} => hello-world : TCP 8080` 

But, if network policy analysis identifies that such connectivity is not permitted, this line will not be added to the report.

adisos:
https://pkg.go.dev/github.com/openshift/api/route/v1#RouteSpec

shireenf-ibm:
- [ ] add support of `Service` objects and map services o pods
- [ ] add support of `Ingress` + tests
- [ ] add support of` Route` + tests

PR #165, opened 2023-06-15
mapping service to its pods 

shireenf-ibm:
#112 
I chose to get service_pods by providing service name and namespace: since In` Route` and `Ingress` objects , services will be provided by their name, and should have same namespace of the object referencing them

### DIFF
--- a/pkg/netpol/eval/internal/k8s/service.go
+++ b/pkg/netpol/eval/internal/k8s/service.go
@@ -15,8 +15,12 @@ package k8s
 
 import (
 	"errors"
+	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 // Service encapsulates k8s service fields that are relevant for connectivity analysis
@@ -43,4 +47,15 @@ func ServiceFromCoreObject(svcObj *corev1.Service) (*Service, error) {
 	}
 
 	return svc, nil
+}
+
+func (svc *Service) ServicSelectorsAsLabelSelector() (labels.Selector, error) {
+	labelsSelector := metav1.LabelSelector{MatchLabels: svc.Selectors}
+	selectorRes, err := metav1.LabelSelectorAsSelector(&labelsSelector)
+	if err != nil {
+		svcStr := types.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}
+		return nil, fmt.Errorf(" service %s : %s", svcStr, selectorErrTitle)
+	}
+
+	return selectorRes, nil
 }
--- a/pkg/netpol/eval/internal/k8s/service.go
+++ b/pkg/netpol/eval/internal/k8s/service.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Service encapsulates k8s service fields that are relevant for connectivity analysis
@@ -36,6 +37,10 @@ func ServiceFromCoreObject(svcObj *corev1.Service) (*Service, error) {
 
 	if svcObj.Spec.Selector == nil {
 		return nil, errors.New("K8s Service without selectors is not supported")
+	}
+
+	if svc.Namespace == "" {
+		svc.Namespace = metav1.NamespaceDefault
 	}
 
 	for k, v := range svcObj.Spec.Selector {
--- a/pkg/netpol/eval/internal/k8s/service.go
+++ b/pkg/netpol/eval/internal/k8s/service.go
@@ -17,7 +17,6 @@ import (
 	"errors"
 
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Service encapsulates k8s service fields that are relevant for connectivity analysis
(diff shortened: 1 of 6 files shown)


pkg/netpol/eval/resources.go
```diff
@@ -343,3 +343,39 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
 	disjointRes := k8s.DisjointIPBlocks(ipbList, []*k8s.IPBlock{newAll})
 	return disjointRes, nil
 }
+
+func (pe *PolicyEngine) getServicePods(svcName, svcNamespace string) ([]*k8s.Pod, error) {
+	svc := pe.getServiceFromServiceNameAndNameSpace(svcName, svcNamespace)
+	// todo: should return error if the service not found?
+	if svc == nil {
+		return nil, fmt.Errorf("service does not exist: %s\\%s", svcNamespace, svcName)
```
adisos, line 351: use `types.NamespacedName` instead of hard-coded str format (check similar usage in other functions)

pkg/netpol/eval/resources.go
```diff
@@ -343,3 +343,39 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
...
+		}
+		if match {
+			res = append(res, pod)
+		}
+	}
+	return res, nil
+}
+
+func (pe *PolicyEngine) getServiceFromServiceNameAndNameSpace(svcName, svcNamespace string) *k8s.Service {
+	for _, svc := range pe.servicesMap {
```
adisos, line 375: consider changing `servicesMap` to be from namespace to service name to service object

pkg/netpol/eval/resources.go
```diff
@@ -343,3 +343,39 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
...
+			val, ok := pod.Labels[k]
+			// no match if the pod doesn't have a lebel with same key, or it has a similar key but with different value
+			if !ok || (ok && val != v) {
+				match = false
+				break // continue to next pod
+			}
+		}
+		if match {
+			res = append(res, pod)
+		}
```
adisos, line 369: matching can be done simpler.. take a look at func `Matches(Labels)` of `Selector` interface at `k8s.io\apimachinery` (pkg\labels\selector.go).  It is used in `pkg\netpol\eval\internal\k8s\netpol.go`.  (`func (np *NetworkPolicy) Selects` )

adisos (changes requested):

pkg/netpol/eval/resources.go
shireenf-ibm: the selector of a service is a `map[string]string` and not of type `Labels.Selector`, so I can not use `Matches` on it

adisos: you can use ` metav1.LabelSelectorAsSelector` (check how it is used in our netpol.go ).
It takes as input `LabelSelector` which has a field named `MatchLabels`  of type `map[string]string`

pkg/netpol/eval/internal/k8s/service.go
```diff
@@ -38,6 +39,10 @@ func ServiceFromCoreObject(svcObj *corev1.Service) (*Service, error) {
 		return nil, errors.New("K8s Service without selectors is not supported")
 	}
 
+	if svc.Namespace == "" {
+		svc.Namespace = metav1.NamespaceDefault
+	}
+
```
adisos: doesn't `scan` pkg assigns `default` namespace for resources without a namespace?

pkg/netpol/eval/resources.go
```diff
@@ -343,3 +349,38 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
...
+	}
+	return res, nil
+}
+
+func (pe *PolicyEngine) getServiceFromServiceNameAndNameSpace(svcName, svcNamespace string) *k8s.Service {
```
adisos, line 381:
```suggestion
func (pe *PolicyEngine) getServiceFromServiceNameAndNamespace(svcName, svcNamespace string) *k8s.Service {
```

pkg/netpol/eval/internal/k8s/service.go
shireenf-ibm: True

pkg/netpol/eval/internal/k8s/service.go
```diff
@@ -14,27 +14,48 @@
...
 
 	return svc, nil
 }
+
+func (svc *Service) ServicSelectorsAsLabelSelector() (labels.Selector, error) {
+	labelsSelector := metav1.LabelSelector{MatchLabels: svc.Selectors}
+	selectorRes, err := metav1.LabelSelectorAsSelector(&labelsSelector)
+	if err != nil {
+		svcStr := types.NamespacedName{Namespace: svc.Namespace, Name: svc.Name}
+		return nil, fmt.Errorf(" service %s : %s", svcStr, selectorErrTitle)
```
adisos, line 57: what is `selectorErrTitle`?

pkg/netpol/eval/service_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/np-guard/netpol-analyzer/pkg/netpol/internal/testutils"
+)
+
+type serviceMapping struct {
+	serviceName      string
+	serviceNamespace string
+	numPods          int
+}
+
+func TestServiceMappingToPods(t *testing.T) {
```
adisos, line 18: add cases with a service that does not exist (with cases for existing and non existing namespace), and a case for a service that has no pods.

pkg/netpol/eval/internal/k8s/service.go
shireenf-ibm: a `const` declared in the `netpol.go `(same pkg) 

`selectorErrTitle  = "selector error"`

adisos (approved):

PR #169, opened 2023-06-20
support scaning Route and Ingress objects

shireenf-ibm:
#112 
task : support scanning (and parsing) Route and k8s Ingress yamls

### DIFF
--- a/pkg/netpol/scan/scan.go
+++ b/pkg/netpol/scan/scan.go
@@ -253,28 +253,21 @@ const (
 	Ingress               string = "Ingress"
 )
 
-// the k8s kinds that scan pkg supports, without List kinds
-var singleResourceK8sKinds = map[string]struct{}{
-	Pod:                   {},
-	Networkpolicy:         {},
-	Namespace:             {},
-	ReplicaSet:            {},
-	Deployment:            {},
-	Statefulset:           {},
-	Daemonset:             {},
-	Job:                   {},
-	CronJob:               {},
-	ReplicationController: {},
-	Service:               {},
-	Route:                 {},
-	Ingress:               {},
-}
-
-// the kinds that scan considers as networking policies objects
-var policyelements = map[string]struct{}{
-	Networkpolicy: {},
-	Route:         {},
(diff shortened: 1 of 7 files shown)


pkg/netpol/scan/scan.go
```diff
@@ -249,6 +266,15 @@ var singleResourceK8sKinds = map[string]struct{}{
...
 	Service:               {},
+	Route:                 {},
+	Ingress:               {},
+}
+
+// the kinds that scan considers as networking policies objects
+var policyelements = map[string]struct{}{
+	Networkpolicy: {},
+	Route:         {},
+	Ingress:       {},
```
adisos, line 277: I don't think Ingress and Route should be considered as policy elements, as I would still like to see a warning if the input dir for the analysis has no network policies. maybe this is another category of elements. consider how to rewrite the function `hasWorkloadsOrNetworkPolicies` as well (it assumes only netpol or workload resources ).

pkg/netpol/scan/scan_test.go
```diff
@@ -175,10 +183,3 @@ func TestNoK8sWorkloadResourcesFoundError(t *testing.T) {
 	noK8sWorkloadResourcesFound := &NoK8sWorkloadResourcesFoundError{}
 	require.True(t, errors.As(errs[0].Error(), &noK8sWorkloadResourcesFound))
 }
-
-func TestK8sWithOpenShiftResources(t *testing.T) {
-	dirPath := filepath.Join(testutils.GetTestsDir(), "acs_security_frontend_demos")
-	objs, errs := scanner.FilesToObjectsList(dirPath)
-	require.Empty(t, errs)
-	require.Len(t, objs, 7) // two deployments, 3 netpols, 2 services (skipping routes and configMaps)
-}
```
adisos, line 184: I think we should keep this test, just try to find another openshift resources that is skipped, now that Route is not skipped.

adisos (changes requested):

adisos (approved):

PR #172, opened 2023-07-05
ingress analyze pkg with route resource

shireenf-ibm:
#112 
task:  Add an Ingress analyzer package with an abstract Ingress src-peer (Ingress Controller pod), and Upsert Route

notes : (some comments are in code too)
* new package: `ingressanalyzer `  (added in a new directory under `eval`, otherwise, will not be able to use `k8s` package which is an internal package of eval)
* currently `ingressanalyzer struct ` supports only `routes map` (k8s ingress map will be added in next PR)
* didn't include the changes to `connlist.go` (`connlist `output) yet (prefer to add it in another PR as the `connlist .go` is in different package)


pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,146 @@
...
+
+type (
+	IngressAnalyzer struct {
+		routesMap map[string]map[string]*k8s.Route // map from namespace to map from route name to its object
+		// k8sIngressMap map[string]map[string]*k8s.Ingress // map from namespace to map from ingress name to its object
+	}
+
+	// NotificationTarget defines an interface for updating the state needed for ingress
+	// decisions
+	NotificationTarget interface {
```
shireenf-ibm, line 34: In order to avoid code duplication (with `PolicyEngine`), we may move this to an eval "utility" package to be used in both structs. (along with general implementations of the `Upsert ` and `Delete` funcs)

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if err != nil {
+			return nil, err
+		}
+	}
+	return ia, err
+}
+
+// GetIngressControllerPod creates a new ingress-controller pod, as the source of all ingress connections
+func (ia *IngressAnalyzer) GetIngressControllerPod() eval.Peer {
```
shireenf-ibm, line 65: this func to be called when computing the `connection object` in `connlist.getConnectionsList()` (for getting the src element of the connection)

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,146 @@
...
+			delete(ia.routesMap, rt.Namespace)
+		}
+	}
+	return nil
+}
+
+////////////////////////////////////////////////////////////////////////////////////
+
+// AllowedIngressConnectionsToAWorkloadPeer return the allowed external ingress-controller's connections to the dst peer
+func (ia *IngressAnalyzer) AllowedIngressConnectionsToAWorkloadPeer(dst eval.Peer, pe *eval.PolicyEngine) (eval.Connection, error) {
```
shireenf-ibm, line 123: - used in the `ingress_analyzer_test.go`
- to be used also in `connlist.getConnectionsList()`
- gets a `Peer` as a parameter to be compatible with the idea of `AllAllowedConnectionsBetweenWorkloadPeers` and for cases of focus-workload used

pkg/netpol/eval/ingressanalyzer/ingress_analyzer_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/np-guard/netpol-analyzer/pkg/netpol/logger"
+	"github.com/np-guard/netpol-analyzer/pkg/netpol/scan"
+)
+
+// global scanner object for testing
+var scanner = scan.NewResourcesScanner(logger.NewDefaultLogger(), false, filepath.WalkDir)
+
+func TestIngressAnalyzerWithRoutes(t *testing.T) {
+	routesNamespace := "frontend"
+	path := filepath.Join(testutils.GetTestsDirFromInternalSubDir(), "acs_security_frontend_demos")
```
shireenf-ibm, line 21: `GetTestsDirFromInternalSubDir()` this was added in `../internal/testutils/utils.go` since the regular `GetTestsDir()` supports only 3 level-ups while more is needed here.
if we want to keep the directory of `ingressanalyzer` we may update the `GetTestsDir()` to support a given level-ups

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,188 @@
...
+		routesMap:   make(map[string]map[string]*k8s.Route),
+	}
+}
+
+// NewIngressAnalyzerWithObjects returns a new IngressAnalyzer with relevant objects
+func NewIngressAnalyzerWithObjects(objects []scan.K8sObject) (*IngressAnalyzer, error) {
+	ia := NewIngressAnalyzer()
+	var err error
+	ia.pe, err = eval.NewPolicyEngineWithObjects(objects)
+	if err != nil {
```
adisos, line 55: do we need to create a new policy engine, or assume it was created and just receive it as input arg? 
I think the connlist will have to create both a policy engine and an IngressAnalyzer that can use the already created policy-engine object. (it will also have to add the ingress-controller pod to get policy-based connectivity between this pod to existing pods in the cluster).

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// GetIngressControllerPod creates a new ingress-controller pod, as the source of all ingress connections
+func (ia *IngressAnalyzer) GetIngressControllerPod() eval.Peer {
+	ingressPod := &k8s.Pod{
+		Name:      "ingress-controller",
+		Namespace: "",
+	}
+	return &k8s.WorkloadPeer{Pod: ingressPod}
+}
```
adisos, line 79: Add a const for the hard-coded fake name (and also add namespace).
Will also have to add this pod to the policy-engine, to evaluate permitted connectivity given an "unknwon" ingress controller pod.

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,188 @@
...
+
+// ClearResources: deletes all current ingress resources
+func (ia *IngressAnalyzer) ClearResources() {
+	ia.pe = nil // or should call pe.ClearResources?
+	ia.routesMap = make(map[string]map[string]*k8s.Route)
+	ia.servicesMap = make(map[string]map[string]*k8s.Service)
+}
```
adisos, line 86: is this function needed?

adisos (changes requested):

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: I started with the approach sending the `PolicyEngine pointer` as a parameter to the `NewIngressAnalyzerWithObjects`,
but changed to this before committing, thinking that we may build only the `IngressAnalyzer ` explicitly in the connlist (the policyEngine will be built anyway - since both Constructors use the same objects list) - and then we can refer to the "ia.pe" to compute/ use the policyEngine.

However, I agree it is more readable to build both explicitly (committed a change)

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: no known use-case -  removed

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -211,3 +211,21 @@ func variantFromLabelsMap(labels map[string]string) string {
 func getFakePodIP() string {
 	return scan.IPv4LoopbackAddr
 }
+
+func (pod *Pod) AllowedConnectionsToPod() ConnectionSet {
+	if len(pod.Ports) == 0 {
+		return MakeConnectionSet(true)
+	}
```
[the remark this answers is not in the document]
shireenf-ibm: should not - deleted

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: > Will also have to add this pod to the policy-engine, to evaluate permitted connectivity given an "unknwon" ingress controller pod.

do we already have a use case for this? should it be included in this PR ?

adisos: not in this PR.. but in the next one

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	for _, rt := range rtMap {
+		if ia.CheckServiceSelectsPod(rt.TargetSvc, peerNs, dstPodPeer.Pod) {
+			return eval.GetConnectionObject((dstPodPeer.Pod).AllowedConnectionsToPod()), nil
+		}
+	}
+
+	return nil, nil // did not find any defined ingress connection to the dst peer
+}
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/route.go
```diff
@@ -0,0 +1,51 @@
...
+	ocroutev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Route encapsulates openshift route fields that are relevant for ingress traffic analysis
+type Route struct {
+	Name      string
+	Namespace string
+	TargetSvc string // service name
+	// todo : should support alternateBackends ? with/without weights (rr)?
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -156,28 +140,44 @@ func (ia *IngressAnalyzer) getServiceFromServiceNameAndNamespace(svcName, svcNam
...
+	return res
+}
 
-	peerNs := dstPodPeer.Namespace()
-	rtMap, ok := ia.routesMap[peerNs]
-	if !ok {
-		return nil, nil // no ingress objects in the pod's namespace
-	}
+// AllowedIngressConnections returns map of the allowed external ingress-controller's connections of each targeted peer
+func (ia *IngressAnalyzer) AllowedIngressConnections() map[*k8s.Pod]eval.Connection {
```
adisos, line 156: The exported function returns a map that its key is of type which is internal.. also, is it safe that the map key is pointer?
should consider using the exported type eval.Peer ? or name string (in level of workload or pod)?

pkg/netpol/eval/ingressanalyzer/ingress_analyzer.go
```diff
@@ -156,28 +140,44 @@ func (ia *IngressAnalyzer) getServiceFromServiceNameAndNamespace(svcName, svcNam
...
+		// if there are no services in same namespace of the route, the routes in this ns will be skipped
+		if _, ok := ia.servicesMap[ns]; !ok {
+			continue
+		}
+		for _, route := range rtMap {
+			routeTargetPods := ia.getRouteTargetedPods(ns, route.TargetServices)
+			// avoid dups in the targetedPodsSet
+			for _, pod := range routeTargetPods {
+				if _, ok := targetedPodsSet[pod]; !ok {
+					targetedPodsSet[pod] = struct{}{}
```
adisos, line 172: can simplify by using a map to bool (val true marks insertion to map), and then check existence by:
`if targetedPodsSet[pod]` (for entry not existing in the map it returns the zero value which is false in the case of bool).

adisos (changes requested):
hi, added a few more comments.
additional question: is it possible to move the new package outside eval?

shireenf-ibm:
> additional question: is it possible to move the new package outside eval?

`ingressanalyzer `package uses the `k8s `package which is internal under `eval`, according to [this ](https://go.dev/doc/go1.4#internalpackages), we can't use `k8s `package from outside `eval `(unless we remove the "internal" directory)

adisos:
> > additional question: is it possible to move the new package outside eval?
> 
> `ingressanalyzer `package uses the `k8s `package which is internal under `eval`, according to [this ](https://go.dev/doc/go1.4#internalpackages), we can't use `k8s `package from outside `eval `(unless we remove the "internal" directory)

right, I meant whether you can make changes so that anything that depends on the internal type is instead exposed by its external interface?

shireenf-ibm:
> > > additional question: is it possible to move the new package outside eval?
> > 
> > 
> > `ingressanalyzer `package uses the `k8s `package which is internal under `eval`, according to [this ](https://go.dev/doc/go1.4#internalpackages), we can't use `k8s `package from outside `eval `(unless we remove the "internal" directory)
> 
> right, I meant whether you can make changes so that anything that depends on the internal type is instead exposed by its external interface?

we still have to use the k8s.Route (and next we'll add the k8s.Ingress object) - (would it be efficient to build interface/s for them?)

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm, line 1: kept the fake ingress-controller "outside" this PR , to be added by the policyEngine in the next PR

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,222 @@
...
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// services analyze
+
+const (
+	selectorError        = "selector conversion error"
+	missingSelectorError = "error : K8s Service without selectors is not supported"
+)
+
+func (ia *IngressAnalyzer) mapServiceToPeers(svc *corev1.Service) error {
+	// get peers selected by the service selctors
```
adisos, line 78:
```suggestion
	// get peers selected by the service selectors
```

adisos (changes requested):
added some more comments

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,222 @@
...
+
+// NewIngressAnalyzer returns a new IngressAnalyzer with an empty initial state
+func NewIngressAnalyzer() *IngressAnalyzer {
+	return &IngressAnalyzer{
+		logger:              logger.NewDefaultLogger(),
+		pe:                  eval.NewPolicyEngine(),
+		servicesToPeersMap:  make(map[string]map[string][]eval.Peer),
+		routesToServicesMap: make(map[string]map[string][]string),
+	}
+}
```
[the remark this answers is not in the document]
shireenf-ibm: not really, it is used from the other C'tr (with objects) mainly to initialize the struct with empty maps.
deleting it moving the initialization rows to the other C'tr

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+		if podPeer.Namespace() != svc.Namespace {
+			continue
+		}
+		if svcLabelsSelector.Matches(labels.Set(podPeer.Pod.Labels)) {
+			res = append(res, peer)
+		}
+	}
+	return res, nil
+}
```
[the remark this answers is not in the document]
shireenf-ibm: moved the relevant code to a new pe.GetSelectedPeers() func.

> I also do not understand why you used `ConvertWorkloadPeerToPodPeer` ? shouldn't the peers be in level of workloads?

In order to check the matching selectors we need to access the `Pod.Labels` of the WorkloadPeer.
`WorkloadPeer` struct does not have access to it `Pod` member as it implements only the `eval.Peer` interface
however, `PodPeer` implements both `eval.Peer` and `k8s.Peer` interfaces , and has an access to the `Pod` element (so the convert was needed)

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	svcLabelsSelector, err := convertServiceSelctorToLabelSelector(svc.Spec.Selector, svcStr)
+	if err != nil {
+		return nil, err
+	}
+	return ia.pe.GetSelectedPeers(svcLabelsSelector, svc.Namespace)
+}
+
+// utility func
+func convertServiceSelctorToLabelSelector(svcSelect map[string]string, svcStr string) (labels.Selector, error) {
```
adisos, line 100:
```suggestion
func convertServiceSelectorToLabelSelector(svcSelector map[string]string, svcStr string) (labels.Selector, error) {
```

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,200 @@
...
+
+func getRouteServices(rt *ocroutev1.Route) ([]string, error) {
+	routeStr := types.NamespacedName{Namespace: rt.Namespace, Name: rt.Name}.String()
+	// Currently, only 'Service' is allowed as the kind of target that the route is referring to.
+	if rt.Spec.To.Kind != "" && rt.Spec.To.Kind != allowedTargetKind {
+		return nil, errors.New(scan.Route + " " + routeStr + ": " + routeTargetKindErr)
+	}
+	if len(rt.Spec.AlternateBackends) > maxBackendServices {
+		return nil, errors.New(scan.Route + " " + routeStr + ": " + routeBackendsErr)
+	}
```
adisos, line 140: why do we need to check this?

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,200 @@
...
+
+	targetSvcs := make([]string, len(rt.Spec.AlternateBackends)+1)
```
adisos, line 142:
```suggestion
	targetServices := make([]string, len(rt.Spec.AlternateBackends)+1)
```

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,200 @@
...
+	targetedPeersSet := make(map[eval.Peer]bool, 0)
+	for ns, rtSvcMap := range ia.routesToServicesMap {
+		// if there are no services in same namespace of the route, the routes in this ns will be skipped
+		if _, ok := ia.servicesToPeersMap[ns]; !ok {
+			continue
+		}
+
+		for _, svcList := range rtSvcMap {
+			routeTargetPeers := ia.getRouteTargetedPeers(ns, svcList)
+			// avoid dups in the targetedPeersSet
```
adisos, line 171:
```suggestion
			// avoid duplicates in the targetedPeersSet
```

pkg/netpol/ingressanalyzer/ingress_analyzer.go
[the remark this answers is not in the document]
shireenf-ibm: resolved - I need to convert peer to a pointer of k8s.WorkloadPeer in order to access its Pod element

pkg/netpol/eval/resources.go
```diff
@@ -351,32 +323,27 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
...
-	// todo: should return error if the service not found?
-	if svc == nil {
-		svcStr := types.NamespacedName{Namespace: svcNamespace, Name: svcName}
-		return nil, fmt.Errorf("service does not exist: %s", svcStr)
-	}
-	svcLabelsSelect, err := svc.ServicSelectorsAsLabelSelector()
+// GetSelectedPeers returns list of peers in the given namespace which match the given labels selector
+func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace string) ([]Peer, error) {
+	res := make([]Peer, 0)
+	peers, err := pe.GetPeersList()
```
adisos, line 329: instead of calling `pe.GetPeersList()` and then re-iterating the entire peers list in level of workloads, can do a simpler iteration as implemented in `GetPeersList` : iterate only `pe.podsMap` , and for a matching pod add it as a workload to the result.

adisos (changes requested):

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: according to [this](https://pkg.go.dev/github.com/openshift/api/route/v1#RouteSpec) - 

> // alternateBackends allows up to 3 additional backends to be assigned to the route.

I assumed we need to check it

pkg/netpol/eval/resources.go
adisos: fix to avoid duplicate workloads in the result

pkg/netpol/ingressanalyzer/ingress_analyzer.go
adisos: can remove this check

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+	}
+
+	// compute allowed conns of each peer pod
+	res := make(map[string]eval.Connection)
+	for peer := range targetedPeersSet {
+		peerStr := types.NamespacedName{Name: peer.Name(), Namespace: peer.Namespace()}.String()
+		peerPod, _ := ia.pe.ConvertWorkloadPeerToPodPeer(peer) // should not get error since all peers in ia.servicesToPeersMap are confirmed
+		res[peerStr] = eval.GetConnectionObject(peerPod.GetAllowedConnectionsToPod())
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: done

pkg/netpol/eval/peer.go
```diff
@@ -14,4 +16,7 @@ type Peer interface {
 	String() string
 	// Kind returns a string of the peer kind in case the peer is a pod/workload, else it returns an empty string
 	Kind() string
+	// GetAllowedConnectionsToPod returns ConnectionSet of the allowed connections to the Pod object of the peer
+	// if it is a pod, else returns empty ConnectionSet
+	GetAllowedConnectionsToPod() k8s.ConnectionSet
```
adisos, line 21: can this exposed function return `eval.Connection` instead of `k8s.ConnectionSet` ?

pkg/netpol/eval/resources.go

pkg/netpol/eval/peer.go
shireenf-ibm: since we can't import `eval`  from its internal package `k8s`.
I reverted adding the func of computing the allowed conns to a peer from the `eval.Peer` interface
and added a new `pe.AllowedConnectionsToPeer(peer)` which computes the conectionset of a pod and returns the connection object to the ingress_analyzer

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,195 @@
...
+//////////////////////////////////////////////////////////////////////////////////////////////
+// Ingress allowed conns
+
+// AllowedIngressConnections returns map of the allowed external ingress-controller's connections of each targeted peer
+func (ia *IngressAnalyzer) AllowedIngressConnections() map[string]eval.Connection {
+	// if there is at least one route/ ingress object that targets a service which selects a dst peer,
+	// then we have an ingress conns to the peer
+
+	// get all targeted peer pods
+	targetedPeersSet := make(map[eval.Peer]bool, 0)
```
adisos, line 158: is it safe to use a map where the keys are pointers to workload objects? the function `GetSelectedPeers` returns new objects on each call...

shireenf-ibm: true, I'll make it a map from workload_name to workload

adisos: maybe you can only have one map `map[string]eval.Connection` built for the result, while collecting all target peers.

shireenf-ibm: done

pkg/netpol/eval/check.go
```diff
@@ -350,3 +350,12 @@ func (pe *PolicyEngine) allAllowedConnections(src, dst string) (k8s.ConnectionSe
 	allowedConns, err := pe.allAllowedConnectionsBetweenPeers(srcPeer.(Peer), dstPeer.(Peer))
 	return allowedConns.(*k8sConnectionSetWrapper).ConnectionSet(), err
 }
+
+// AllowedConnectionsToPeer returns the allowed connections to a workload peer
+func (pe *PolicyEngine) AllowedConnectionsToPeer(peer Peer) Connection {
```
adisos, line 355: rename func to `GetPeerExposedProtocolsAndPorts` (also the internal version, for a pod). 
the term "allowed connections" may be confusing

pkg/netpol/eval/check.go
```diff
@@ -350,3 +350,12 @@ func (pe *PolicyEngine) allAllowedConnections(src, dst string) (k8s.ConnectionSe
...
+	workload, ok := peer.(*k8s.WorkloadPeer)
```
adisos, line 356: can use a switch on the type of Peer, to handle also other types that implement the Peer interface

pkg/netpol/eval/resources.go
```diff
@@ -351,32 +328,16 @@ func (pe *PolicyEngine) getDisjointIPBlocks() ([]*k8s.IPBlock, error) {
...
-		return nil, fmt.Errorf("service does not exist: %s", svcStr)
-	}
-	svcLabelsSelect, err := svc.ServicSelectorsAsLabelSelector()
-	if err != nil {
-		return nil, err
-	}
-	res := make([]*k8s.Pod, 0)
-	for _, pod := range pe.podsMap {
-		if pod.Namespace != svcNamespace {
+// GetSelectedPeers returns list of peers in the given namespace which match the given labels selector
```
adisos, line 331:
```suggestion
// GetSelectedPeers returns list of workload peers in the given namespace which match the given labels selector
```

adisos (approved):

Issue #154, opened 2023-06-05
add semantic-diff cmd


adisos:
allow connectivity-diff between two directories containing workloads and netpols manifests.

shireenf-ibm:
planned PRs (on the semantic_diff_command branch):
[] adding diff command (pr #174)
[] adding diff functionality - diff conns on common peers
[] writing diff output 
[] support multiple formats

adisos:
- diff-part1: handle connection equality issue
- diff-part2: let added tests run with unit-tests, review 
- diff-part3: output formats (txt,csv,md)
- diff-part4: dot format 

consider for later:
- merge disjoint-ip-blocks in output where possible
- mark peers which are removed/added

adisos:
for the case of an added/removed workload, let's add to the diff-type that info.
for example, in the current report if we have `src: A, dst: B, dir1: NoConn, dir2: TCP, diff-type: added` , and if `B` is a new workload that does not exist in dir1, then the report should be with `diff-type: added (workload A added)`

PR #174, opened 2023-07-10
Semantic diff

shireenf-ibm:
first commit : very initial commit - adding `diff `command with `--other` flag but with empty functionality
#154


shireenf-ibm:
hi @adisos, added new commit with first test , changes explained in the commit message

tests/onlineboutique_workloads_changed_netpols/diff_output_from_onlineboutique_workloads
adisos, line 1: let's change the output as follows:

`source: <peer name> , destination: <peer name> ,  connections before: <conn1> , connections after: <conn2>`

and let's have all lines at this format, without separation to lost/new/changed connections.
only header of changed connections.
(removed connections are also changed, only with "no conns" at connections-after)

adisos (changes requested):

pkg/netpol/diff/diff.go
```diff
@@ -79,31 +79,42 @@ func (d diffMap) update(key string, isFirst bool, c connlist.Peer2PeerConnection
...
 
 	return res, nil
 }
 
+// checks whether two connlist.Peer2PeerConnection objects are equal
+func equalConns(firstConn, secondConn connlist.Peer2PeerConnection) bool {
+	return firstConn.AllProtocolsAndPorts() == secondConn.AllProtocolsAndPorts() &&
+		reflect.DeepEqual(firstConn.ProtocolsAndPorts(), secondConn.ProtocolsAndPorts())
+}
+
```
adisos, line 117: can you convert to `common.ConnectionSet` objects and use its `Equals` method instead?

shireenf-ibm: it is possible, but we'll need to wait until merging the "ingress_analyzer" PRs , since `common` is not merged into main branch yet

tests/onlineboutique_workloads_changed_netpols/diff_output_from_onlineboutique_workloads
shireenf-ibm: done

adisos: looks good!
few more comments:
- change title to "Connectivity diff:"
- replace "no conns" with "No Connections" 
- replace "before" and "after" with "dir1" and "dir2" , and also align with input args: "dir1" and "dir2" instead of "dirpath" and "other"

adisos: looks good. can you also change the names of the command args accordingly (dir1 and dir2 instead of dirpath and other)

pkg/netpol/diff/diff.go
shireenf-ibm: committed changes to #180.
closing this PR

PR #175, opened 2023-07-12
ingress support in the ingress analyzer

shireenf-ibm:


pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -146,6 +153,49 @@ func getRouteServices(rt *ocroutev1.Route) ([]string, error) {
...
+			// Currently, only IngressBackend with Service is supported
+			return nil, errors.New(scan.Ingress + " " + ingressStr + ": " + defaultBackendError)
+		}
+		backendServices[0] = ing.Spec.DefaultBackend.Service.Name
+	}
+	// add service names from the Ingress rules
+	for _, rule := range ing.Spec.Rules {
+		for _, path := range rule.IngressRuleValue.HTTP.Paths {
+			if path.Backend.Service == nil {
+				return nil, errors.New(scan.Ingress + " " + ingressStr + ": " + ruleBackendError)
```
adisos, line 191: - let's avoid errors, just ignore rules that are not supported or not containing expected values.
- what about specified ports on a certain rule?

shireenf-ibm: > let's avoid errors, just ignore rules that are not supported or not containing expected values.

done

shireenf-ibm: > what about specified ports on a certain rule?

I thought at this version we'll assume the ingress rule with a service has the correct port (the same of the service's port)
will update

adisos: it's also possible that the pod exposes multiple ports and only one of them is specified in the ingress rule.

adisos: please add logger warnings when ignoring certain rules or resources

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -156,29 +203,43 @@ func (ia *IngressAnalyzer) AllowedIngressConnections() map[string]eval.Connectio
...
 			continue
 		}
 
-		for _, svcList := range rtSvcMap {
-			routeTargetPeers := ia.getRouteTargetedPeers(ns, svcList)
+		for _, svcList := range objSvcMap {
+			ingressObjTargetPeers := ia.getIngressObjectTargetedPeers(ns, svcList)
 			// avoid duplicates in the result
-			for _, peer := range routeTargetPeers {
+			for _, peer := range ingressObjTargetPeers {
```
adisos, line 231: to get `peerStr` in the next line, can use `peer.String()` instead of having to use `types.NamespacedName`

shireenf-ibm:
> hi @adisos , thinking about the connections (and while computing k8s-ingress connections) May An ingress (from Route/k8s-Ingress) connection be on a protocol other than TCP? It seems to me no, if I am right should we get the pods' exposed TCP connections for the ingress analyzer , not all protocols and ports?

since k8s Ingress, and routes does not specify the protocol on their ports fields , then default protocol which is TCP is used
I changed the connections to pods returned from ingress-analyzer to be TCP only

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -106,75 +125,151 @@ func convertServiceSelectorToLabelSelector(svcSelector map[string]string, svcStr
 	return selectorRes, nil
 }
 
-// //////////////////////////////////////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////////////////////
```
adisos, line 128: in the function above this line `convertServiceSelectorToLabelSelector` , should return the `err` itself returned from `LabelSelectorAsSelector` (we don't want to miss/override the original error)

adisos (changes requested):
few more comments.
please also add documentation to functions where possible. (`utility func` is not sufficiently informative).

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -189,3 +286,63 @@ func (ia *IngressAnalyzer) getRouteTargetedPeers(ns string, svcList []string) []
...
+func (ia *IngressAnalyzer) getIngressObjectTargetedPeersAndPorts(ns string, svcList []serviceInfo) map[eval.Peer]*common.ConnectionSet {
+	res := make(map[eval.Peer]*common.ConnectionSet)
+	for _, svc := range svcList {
+		peers, ok := ia.servicesToPeersMap[ns][svc.serviceName]
+		if !ok {
+			ia.logger.Warnf("ignoring target service " + svc.serviceName + " : service not found")
+		}
+		for _, peer := range peers {
+			permittedPeerConn := common.MakeConnectionSet(false)
+			// check if its TCP conns contains the reuired port
```
adisos, line 333:
```suggestion
			// check if its TCP conns contains the required port
```

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -189,3 +286,63 @@ func (ia *IngressAnalyzer) getRouteTargetedPeers(ns string, svcList []string) []
...
+			peerTCPConn := eval.GetPeerExposedTCPConnections(peer)
+			if peerTCPConn.Contains(svc.serviceTargetPort.String(), string(corev1.ProtocolTCP)) {
+				permittedPort := common.PortSet{}
+				permittedPort.AddPort(svc.serviceTargetPort)
```
adisos, line 337: we should not keep named ports, but rather convert them to port numbers

adisos (changes requested):

pkg/netpol/ingressanalyzer/service_test.go
adisos, line 1: in `serviceMapping` struct, rename `numPods` to `numWorkloads`

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -65,119 +76,205 @@ func NewIngressAnalyzerWithObjects(objects []scan.K8sObject, pe *eval.PolicyEngi
...
 	ia.servicesToPeersMap[svc.Namespace][svc.Name] = peers
 	return nil
 }
 
+// getServicePeers given a service return its selected peers
 func (ia *IngressAnalyzer) getServicePeers(svc *corev1.Service) ([]eval.Peer, error) {
 	svcStr := types.NamespacedName{Name: svc.Name, Namespace: svc.Namespace}.String()
 	if svc.Spec.Selector == nil {
-		return nil, errors.New(scan.Service + " " + svcStr + " " + missingSelectorError)
+		ia.logger.Warnf("ignoring " + scan.Service + " " + svcStr + ": " + missingSelectorWarning)
```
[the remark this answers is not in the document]
shireenf-ibm: this warning was initially an error - so we had a test -"TestNotSupportedService" , since the warning is raised by the logger now, I didn't check the warning message but that the servicesMap is empty.
Since we don't warn if there are no policies/ policy-rules selecting pod I didn't warn if a service doesn't actually select any peer / if route/Ingress don't select any service (warning raised only if the target resource is not supported , i.e. not a service)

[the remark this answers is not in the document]
shireenf-ibm: > this warning was initially an error - so we had a test -"TestNotSupportedService" , since the warning is raised by the logger now, I didn't check the warning message (testing logger messages is a little bit complicated and I was afraid to have a time waste) but that the servicesMap is empty. Since we don't warn if there are no policies/ policy-rules selecting pod I didn't warn if a service doesn't actually select any peer / if route/Ingress don't select any service (warning raised only if the target resource is not supported , i.e. not a service)

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -189,3 +289,77 @@ func (ia *IngressAnalyzer) getRouteTargetedPeers(ns string, svcList []string) []
...
+				var portNum int
+				if svc.serviceTargetPort.Type == intstr.Int {
+					portNum = svc.serviceTargetPort.IntValue()
+				} else {
+					portInt, err := ia.pe.ConvertPeerNamedPort(svc.serviceTargetPort.StrVal, peer)
+					if err != nil {
+						return nil, err
+					}
+					portNum = int(portInt)
+				}
```
adisos, line 353: do we have a test with k8s Ingress rule with named port ?

shireenf-ibm: added

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -65,119 +77,208 @@ func NewIngressAnalyzerWithObjects(objects []scan.K8sObject, pe *eval.PolicyEngi
...
-func (ia *IngressAnalyzer) AllowedIngressConnections() map[string]eval.Connection {
+func (ia *IngressAnalyzer) AllowedIngressConnections() (map[string]common.Connection, error) {
 	// if there is at least one route/ ingress object that targets a service which selects a dst peer,
 	// then we have ingress connections to that peer
 
 	// get all targeted workload peers and compute allowed conns of each workload peer
-	res := make(map[string]eval.Connection)
-	for ns, rtSvcMap := range ia.routesToServicesMap {
+	// 1. from routes
+	routesResult := ia.allowedIngressConnectionsByRoutes()
```
adisos, line 236: in the routes analysis, why ignoring `targetPort` specified in the resource?

shireenf-ibm: will add it (at first we preferred to not include it)

shireenf-ibm: I see in the examples that the route.targetPort may refer to the service.port.name which may differ from the service.port.targetPort(which eventually is the port in the peer)

so I'll change the servicesToPeersMap from being services -> peers to be services -> {peers , servicePorts}.

adisos (changes requested):

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -332,34 +322,72 @@ func (ia *IngressAnalyzer) getIngressObjectTargetedPeersAndPorts(ns string,
...
 }
+
+// getIngressPeerConnection returns the ingress connection to a peer based on the required port specified in the ingress objects
+func (ia *IngressAnalyzer) getIngressPeerConnection(peer eval.Peer, actualServicePorts []corev1.ServicePort,
+	requiredPort intstr.IntOrString) (*common.ConnectionSet, error) {
+	peerTCPConn := eval.GetPeerExposedTCPConnections(peer)
+	// if the required port is not specified - all peer's TCP connections are allowed for ingress
+	if requiredPort.IntVal == 0 && requiredPort.StrVal == "" {
+		return peerTCPConn, nil
+	}
```
adisos, line 351: doesn't it depend on what the service exposes?

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -332,34 +322,72 @@ func (ia *IngressAnalyzer) getIngressObjectTargetedPeersAndPorts(ns string,
...
+			svcPodAccessPort.IntVal = svcPort.Port
+		}
+		// checking if the service port matches the required port, if yes returning its pod access port
+		switch {
+		case svcPort.Name != "" && svcPort.Name == requiredPort.StrVal, // or
+			svcPort.Port == requiredPort.IntVal, // or
+			svcPort.TargetPort == requiredPort:
+			return svcPodAccessPort
+		}
+	}
```
adisos, line 391: the service target port should match the pod's port

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: true - will change

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: added explaination - copied from [here](https://pkg.go.dev/k8s.io/api@v0.26.1/core/v1#ServicePort.TargetPort)

pkg/netpol/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: updated code

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+			return nil, err
+		}
+	}
+	return ia, err
+}
+
+// IsEmpty returns whether there are no services to consider for Ingress analysis
+func (ia *IngressAnalyzer) IsEmpty() bool {
+	return len(ia.routesToServicesMap) == 0 && len(ia.k8sIngressToServicesMap) == 0
+}
```
adisos, line 95: why is the analysis required if there are Route/Ingress objects, but no Service objects?

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+	if _, ok := ia.servicesToPortsAndPeersMap[svc.Namespace]; !ok {
+		ia.servicesToPortsAndPeersMap[svc.Namespace] = make(map[string]peersAndPorts)
+	}
+	ia.servicesToPortsAndPeersMap[svc.Namespace][svc.Name] = peersAndPorts{peers: peers, ports: svc.Spec.Ports}
+	return nil
+}
+
+// getServicePeers given a service return its selected peers
+func (ia *IngressAnalyzer) getServicePeers(svc *corev1.Service) ([]eval.Peer, error) {
```
adisos, line 121:
```suggestion
func (ia *IngressAnalyzer) getServiceSelectedPeers(svc *corev1.Service) ([]eval.Peer, error) {
```

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+// ///////////////////////////////////////////////////////////////////////////////////////////////
+// k8s Ingress objects analysis
+
+const (
+	defaultBackendWarning = "ignoring default backend"
+	ruleBackendWarning    = "ignoring rule backend without Service"
+)
+
+// mapk8sIngressToServices populates k8sIngressToServicesMap
+func (ia *IngressAnalyzer) mapk8sIngressToServices(ing *netv1.Ingress) {
```
adisos, line 200:
```suggestion
func (ia *IngressAnalyzer) mapK8sIngressToServices(ing *netv1.Ingress) {
```

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+		return
+	}
+	if _, ok := ia.k8sIngressToServicesMap[ing.Namespace]; !ok {
+		ia.k8sIngressToServicesMap[ing.Namespace] = make(map[string][]serviceInfo)
+	}
+	ia.k8sIngressToServicesMap[ing.Namespace][ing.Name] = services
+}
+
+// getk8sIngressServices given k8s-Ingress object returns its targeted services info
+func (ia *IngressAnalyzer) getk8sIngressServices(ing *netv1.Ingress) []serviceInfo {
```
adisos, line 212:
```suggestion
func (ia *IngressAnalyzer) getK8sIngressServices(ing *netv1.Ingress) []serviceInfo {
```

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+//////////////////////////////////////////////////////////////////////////////////////////////
+// Ingress allowed connections
+
+type PeerAndIngressConnSet struct {
+	// used for computations and rtype of allowedIngressConnectionsByResourcesType
+	Peer    eval.Peer
+	ConnSet *common.ConnectionSet
+}
```
adisos, line 256:
```suggestion

// PeerAndIngressConnSet captures Peer object as allowed target from ingress-controller Pod, with its possible connections
type PeerAndIngressConnSet struct {
	Peer    eval.Peer
	ConnSet *common.ConnectionSet
}
```

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,408 @@
...
+
+// AllowedIngressConnections returns a map of the possible connections from ingress-controller pod to workload peers,
+// as inferred from Ingress and Route resources. The map is from a workload name to its connection object.
```
adisos, line 259:
```suggestion
// as inferred from Ingress and Route resources. The map is from a workload name to its PeerAndIngressConnSet object.
```

adisos (changes requested):

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
shireenf-ibm: if there are no service objects, the ia.routesToServicesMap and ia.k8sIngressToServicesMap will be empty. 
it there are no services matching the Route/Ingress object, then it will not have an entry in the relevant map

shireenf-ibm: will update the return statement

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+			return nil, err
+		}
+	}
+	return ia, err
+}
+
+// IsEmpty returns whether there are no services to consider for Ingress analysis
+func (ia *IngressAnalyzer) IsEmpty() bool {
+	return len(ia.servicesToPortsAndPeersMap) == 0 && len(ia.routesToServicesMap) == 0 && len(ia.k8sIngressToServicesMap) == 0
+}
```
adisos, line 95:
```suggestion
	return len(ia.servicesToPortsAndPeersMap) == 0 || ( len(ia.routesToServicesMap) == 0 && len(ia.k8sIngressToServicesMap) ) == 0
}
```

pkg/netpol/connlist/connlist.go
```diff
@@ -383,27 +422,64 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+		if err != nil {
+			return nil, err
+		}
+		peerAndConn.ConnSet.Intersection(peConn.(*common.ConnectionSet))
+		if peerAndConn.ConnSet.IsEmpty() {
+			ca.logger.Warnf("Analysis of Ingress/Route resources inferred there is a configured connectivity from ingress-controller to workload " +
+				peerStr + ", but network policies may be blocking access to this workload" +
+				"(unless specifically permitted to the installed ingress controller pod)")
+			continue
+		}
```
adisos, line 469: change warning as follows:
`Ingress/Route (name) specified workload (name) as a backend, but network policies are  blocking ingress connections from an arbitrary in-cluster source to this workload. Connectivity map will not include a possibly allowed connection between the ingress controller and this workload.`

replace `(name)` with the relevant resource name

shireenf-ibm: updated, however
we don't have an indication from here of the ingress/route name/s which selects the workload (+ more than one ingress/route objects may select same service (mapped to same peer)).

adisos: can you collect this information to be displayed in this message?

shireenf-ibm: yes its possible, updating

adisos: let's be consistent with resource naming in the same format

shireenf-ibm:
hi @adisos , added tests with all formats
rewritten the` TestConnList func` to be table-driven, uses the `testify/require pkg` and support  multiple formats
(deleted tests with specific fromats as they are now contained in TestConnList )

adisos (approved):

PR #177, opened 2023-07-16
Adding ingress analysis to connlist pkg

shireenf-ibm:
this branch was created from #175 's branch so we may merge it only after merging the previous PR

added tests with expected results
ingress results printed are considering both pe.policies and ingress connections

for example, in the `acs_security_frontend_demos` we have routes with target services (to two deployments) but, also we have a policy that denies all ingress to these deployments , so we get that there are no ingress connections (currently, printed the ingress results with the `No Connections`)


pkg/netpol/connlist/connlist.go
```diff
@@ -362,7 +368,7 @@ func (ca *ConnlistAnalyzer) includePairOfWorkloads(src, dst eval.Peer) bool {
 }
 
 // getConnectionsList returns connections list from PolicyEngine object
-func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2PeerConnection, error) {
+func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine, ia *ingressanalyzer.IngressAnalyzer) ([]Peer2PeerConnection, error) {
```
adisos, line 371: can separate the code to two functions called by this function, so that the first has the original code and the other has the part that you added?

pkg/netpol/connlist/connlist.go
```diff
@@ -400,6 +407,38 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
 			res = append(res, connectionObj)
 		}
 	}
+
+	// analyze ingress connections - create connection objects for relevant ingress analyzer connections
+	if !ia.IsEmpty() {
```
adisos, line 412: avoid having most of the code in the inner indentation, use the negation of the current condition

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -73,6 +78,11 @@ func NewIngressAnalyzerWithObjects(objects []scan.K8sObject, pe *eval.PolicyEngi
 	return ia, err
 }
 
+// IsEmpty returns weither the ingress analyzer is empty
```
adisos, line 81:
```suggestion
// IsEmpty returns whether there are no services to consider for Ingress analysis
```

pkg/netpol/connlist/connlist.go
```diff
@@ -400,6 +407,38 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+				continue
+			}
+			// otherwise, compute allowed connections based on pe.policies to the peer, then intersect the conns with
+			// ingress connections to the peer -> the intersection will be appended to the result
+			peConn, err := pe.AllAllowedConnectionsBetweenWorkloadPeers(ingressControllerPod, peer)
+			if err != nil {
+				ca.errors = append(ca.errors, newResourceEvaluationError(err))
+				return nil, err
+			}
+			ingressConns[peer.String()].ConnectionsIntersection(peConn)
```
adisos, line 432: compute the intersection by converting the connection objects to `*common.ConnectionSet`

pkg/netpol/connlist/connlist.go
```diff
@@ -400,6 +407,38 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+			connectionObj := &connection{
```
adisos, line 433:
```suggestion
			p2pConnection := &connection{
```

pkg/netpol/connlist/connlist.go
```diff
@@ -400,6 +407,38 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+				src:               ingressControllerPod,
+				dst:               peer,
+				allConnections:    ingressConns[peer.String()].AllConnections(),
+				protocolsAndPorts: ingressConns[peer.String()].ProtocolsAndPortsMap(),
+			}
+			res = append(res, connectionObj)
```
adisos, line 439:
```suggestion
			res = append(res, p2pConnection)
```

adisos: do we want to keep in the output a line for `no connection` case? 
or instead issue a warning about a connection specified by ingress/route resource that is blocked by network policies?

pkg/netpol/eval/resources.go
```diff
@@ -339,3 +346,17 @@ func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace st
 	}
 	return res
 }
+
+func (pe *PolicyEngine) AddPodByNameAndNamespace(name, ns string) (Peer, error) {
```
adisos, line 350: exposed functions need to have documentation

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -30,6 +30,11 @@ import (
 	"github.com/np-guard/netpol-analyzer/pkg/netpol/scan"
 )
 
+const (
+	IngressPodName = "ingress-controller"
+	IngressPodNs   = "ingress-controller-ns"
```
adisos, line 35:
```suggestion
	IngressPodNamespace   = "ingress-controller-ns"
```

pkg/netpol/ingressanalyzer/ingress_analyzer.go
```diff
@@ -30,6 +30,11 @@ import (
 	"github.com/np-guard/netpol-analyzer/pkg/netpol/scan"
 )
 
+const (
```
adisos, line 33:
```suggestion
const (
//  The actual ingress controller pod is usually unknown and not available in the input resources for the analysis. 
// IngressPodName and IngressPodNamespace are used to represent  that pod with those placeholder values for name and namespace.
```

adisos (changes requested):
added a few comments.
still need to take a look at examples and tests.

adisos:
> this branch was created from #175 's branch so we may merge it only after merging the previous PR
> 
> added tests with expected results ingress results printed are considering both pe.policies and ingress connections
> 
> for example, in the `acs_security_frontend_demos` we have routes with target services (to two deployments) but, also we have a policy that denies all ingress to these deployments , so we get that there are no ingress connections (currently, printed the ingress results with the `No Connections`)

I changed the base branch so that it only shows the changes with respect to that branch, and not with respect to main.

pkg/netpol/connlist/connlist.go
```diff
@@ -375,6 +381,32 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
 		return nil, err
 	}
 
+	// compute connections between peers based on pe policies rules
```
adisos, line 384:
```suggestion
	// compute connections between peers based on pe analysis of network policies
```

pkg/netpol/connlist/connlist.go
```diff
@@ -384,21 +416,60 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+	ingressConns, err := ia.AllowedIngressConnections()
+	if err != nil {
+		return nil, err
+	}
+	// adding the ingress controller pod to the policy engine,
+	ingressControllerPod, err := pe.AddPodByNameAndNamespace(ingressanalyzer.IngressPodName, ingressanalyzer.IngressPodNamespace)
+	if err != nil {
+		return nil, err
+	}
+	for _, peer := range peerList {
```
adisos, line 450: no need to iterate the peers list if the map `ingressConns` is empty (for example no Ingress or Route resources were configured)

adisos: it might be more efficient to iterate only the peers from `ingressConns` map. since the keys there are the string of the peer and not the peer object itself, you can modify the returned map `ingressConns` to be from peer string to a new type (struct) that holds both ConnectionSet and the Peer object itself.

pkg/netpol/connlist/connlist.go
```diff
@@ -384,21 +416,60 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+		}
+		// otherwise, compute allowed connections based on pe.policies to the peer, then intersect the conns with
+		// ingress connections to the peer -> the intersection will be appended to the result
+		peConn, err := pe.AllAllowedConnectionsBetweenWorkloadPeers(ingressControllerPod, peer)
+		if err != nil {
+			return nil, err
+		}
+		ingressConns[peer.String()].(*common.ConnectionSet).Intersection(peConn.(*common.ConnectionSet))
+		if ingressConns[peer.String()].IsEmpty() {
+			ca.logger.Warnf("Ingrees Connections to: " + peer.String() + " were blocked by network-policies")
```
adisos, line 463: change the warning string: "Analysis of Ingress/Route resources inferred there is a configured connectivity from ingress-controller to workload [name], but network policies may be blocking access to this workload (unless specifically permitted to the installed ingress controller pod)"

pkg/netpol/connlist/connlist.go
```diff
@@ -375,6 +381,32 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine) ([]Peer2Pe
...
+	peersAllowedConns, err := ca.getConnectionsBetweenPeers(pe, peerList)
+	if err != nil {
+		ca.errors = append(ca.errors, newResourceEvaluationError(err))
+		return nil, err
+	}
+	res = append(res, peersAllowedConns...)
+
+	if ia.IsEmpty() {
+		return res, nil
+	}
```
adisos, line 394: there could be services without any configured Ingress/Route resources, in which case we also do not need to add the ingress analysis..

pkg/netpol/connlist/connlist_test.go
adisos, line 1: tests issues:
on ` tests/acs_security_frontend_demos/` , the output is:
```
2023/07/19 10:37:54 in file: tests\acs_security_frontend_demos\asset_cache_deployment.yaml, document: 0, skipping object with type: ServiceAccount
2023/07/19 10:37:54 in file: tests\acs_security_frontend_demos\webapp_configmap.yaml, document: 0, skipping object with type: ConfigMap
2023/07/19 10:37:54 Ingrees Connections to: frontend/webapp[Deployment] were blocked by network-policies
2023/07/19 10:37:54 Ingrees Connections to: frontend/asset-cache[Deployment] were blocked by network-policies


```
I think that if the section of the connectivity lines is empty, it should be specified clearly in the output, something like "connectivity analysis found no allowed connectivity between pairs from the configured workloads or external IP-blocks"

adisos: additional test issue:
on `tests/bad_netpols` , the output is:
```
Error: network policy default/shippingservice-netpol named port error: cannot convert named port for an IP destination
Usage:
  k8snetpolicy list [flags]

Examples:
  # Get list of allowed connections from resources dir path
  k8snetpolicy list --dirpath ./resources_dir/

  # Get list of allowed connections from live k8s cluster
  k8snetpolicy list -k ./kube/config

Flags:
      --focusworkload string   Focus connections of specified workload name in the output
  -h, --help                   help for list
  -o, --output string          Required output format (txt,json,dot,csv,md) (default "txt")

Global Flags:
  -c, --context string      Kubernetes context to use when evaluating connections in a live cluster
      --dirpath string      Resources dir path when evaluating connections from a dir
  -k, --kubeconfig string   Path and file to use for kubeconfig when evaluating connections in a live cluster 

```

why do we get usage info when the command was valid? (used `./bin/k8snetpolicy list --dirpath tests/bad_netpols` )?

pkg/netpol/connlist/connlist.go
shireenf-ibm: changed the ia.IsEmpty to check the ingress and routes map (this func will not be called if the ingress-analyzer is empty).
updated the loop to iterate only on the ingress connections

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: > I think that if the section of the connectivity lines is empty, it should be specified clearly in the output, something like "connectivity analysis found no allowed connectivity between pairs from the configured workloads or external IP-blocks" (maybe as warning from the logger)

done

shireenf-ibm: > additional test issue: on `tests/bad_netpols` , the output is:
>
> why do we get usage info when the command was valid? (used `./bin/k8snetpolicy list --dirpath tests/bad_netpols` )?

since this issue occurs on main branch too.
I will resolve in a separate PR

shireenf-ibm:
hi @adisos , merging this PR with the base one. ( all comments resolved except the `usage` message)

PR #182, opened 2023-07-24
diff - part 3

shireenf-ibm:
csv,md,txt formats + diff errors handling


shireenf-ibm:
in the next commit will add error handling support

PR #195, opened 2023-07-31
Diff code updates

shireenf-ibm:


pkg/netpol/connlist/connlist_test.go
```diff
@@ -7,19 +7,15 @@ import (
...
 	"github.com/stretchr/testify/require"
 )
 
 func getConnlistFromDirPathRes(stopOnErr bool, path string) (*ConnlistAnalyzer, []Peer2PeerConnection, error) {
-	var analyzer *ConnlistAnalyzer
-	if stopOnErr {
-		analyzer = NewConnlistAnalyzer(WithStopOnError())
-	} else {
-		analyzer = NewConnlistAnalyzer()
-	}
```
adisos, line 22: let's keep `WithStopOnError()` as it was before, I think we should avoid changing this api interface, as it makes sense to be called with no args (when this option is required).

pkg/netpol/diff/diff.go
```diff
@@ -51,9 +52,9 @@ func WithOutputFormat(outputFormat string) DiffAnalyzerOption {
 
 // WithStopOnError is a functional option which directs DiffAnalyzer to stop any processing after the
 // first severe error.
-func WithStopOnError() DiffAnalyzerOption {
+func WithStopOnError(stopOnError bool) DiffAnalyzerOption {
```
adisos, line 55: same here, let's match to  original api

pkg/netpol/common/outputFormats.go
```diff
@@ -0,0 +1,10 @@
+package common
+
+const (
```
adisos, line 3:
```suggestion
// formats supported for output of various commands
const (
```

PR #196, opened 2023-08-01
diff fatal errors examples 

shireenf-ibm:
wrong format / fatal errors returned from ca


adisos (changes requested):
also add errors in warning level (as in connlis), and add relevant tests (e.g. no network policies in dir)

adisos:
shouldn't the base branch be `semantic_diff_cmd` ?

shireenf-ibm:
> shouldn't the base branch be `semantic_diff_cmd` ?

it should be , locally, I started each branch from the previous one (not from semantic_diff_cmd) , so to avoid seeing all commits from other branches I chose the base same as the local base

pkg/netpol/diff/diff_errors.go
adisos, line 1: move the definition of `type DiffError interface` to this file

pkg/netpol/diff/diff_errors.go
```diff
@@ -2,7 +2,9 @@ package diff
 
 // diffGeneratingError - DiffError that may arise while producing the connectivity diff report
 type diffGeneratingError struct {
```
adisos, line 4: why do the functions `func (e *diffGeneratingError) IsFatal() ` and `func (e *diffGeneratingError) IsSevere()` return constant values and not depend on the values of `fatal, severe` members of `diffGeneratingError` ?

pkg/netpol/diff/diff.go
```diff
@@ -91,13 +91,17 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
...
 		return nil, err
 	}
 	if conns2, err = caAnalyzer.ConnlistFromDirPath(dirPath2); err != nil {
-		da.errors = append(da.errors, newConnectionsAnalyzingError(err))
+		da.errors = append(da.errors, newConnectionsAnalyzingError(err, true, false))
 		return nil, err
 	}
+	// appending connlist warnings and severe errors to diff_errors
+	for _, e := range caAnalyzer.Errors() {
+		da.errors = append(da.errors, newConnectionsAnalyzingError(e.Error(), e.IsFatal(), e.IsSevere()))
```
adisos, line 103: why not just append `e` itself?

pkg/netpol/diff/diff_test.go
```diff
@@ -75,14 +63,150 @@ func TestDiff(t *testing.T) {
...
+	isCaOtherErr    bool // not fatal
+	isFormattingErr bool
+	format          string
+}
+
+var caErrType = &connectionsAnalyzingError{}     // error returned from a func on the ConnlistAnalyzer object
+var formattingErrType = &resultFormattingError{} // error returned from getting/writing output format
+
+func TestDiffErrors(t *testing.T) {
+	// following tests will be run with stopOnError, testing err string and diff err type
```
adisos, line 89: don't we want to test with and without `stopOnError`? (not only with this option enabled)

pkg/netpol/diff/diff.go
shireenf-ibm: oops! fixed

pkg/netpol/diff/diff_test.go
shireenf-ibm: updated

pkg/netpol/diff/diff.go
adisos: why not append as `append(da.errors, e)` ?

PR #197, opened 2023-08-01
Diff tests based nca

shireenf-ibm:
some tests won't be imported in this PR , description in #200


shireenf-ibm:
some tests won't be imported in this PR , description in #200

adisos (commented):
consider adding also connlist tests for these added tests

shireenf-ibm:
Hi @adisos , after analyzing the tests in #200, 
turns out that most of the tests that were added here have pods with same `ownerReferences name` but different `labels` in their pods lists. (e.g. tests with pods_list.json/yaml, bank_pod_list.json/yaml_version.yml) 
Those tests PASS and get same results as NCA since the policies/ policy-rules selectors are not selecting pods with the "problematic" labels .

- So, should we add the warning message for all of these tests?
and maybe an error/ stop running if the policy selectors causing a problem? (in this case we may get different result for each run)

adisos (approved):

PR #201, opened 2023-08-01
issue #163: support NetworkPolicyList resources

adisos:


tests/acs-security-demos-with-netpol-list/expected_netpol_output.yaml
shireenf-ibm, line 1: I would suggest to change this file name to something like : `netpol_list.yaml` or `acs_netpollist.yaml`

adisos: done

PR #202, opened 2023-08-02
Diff with added/removed peers info

shireenf-ibm:


shireenf-ibm:
note: 
in current ,  have src: A, dst: B, dir1: NoConn, dir2: TCP, diff-type: added , and if both A and B are new workloads then the report will with diff-type: added (workload A added)(workload B added)

tests/acs-security-demos-no-routes/diff_output_from_acs-security-demos.txt
```diff
@@ -1,3 +1,3 @@
 Connectivity diff:
-source: {ingress-controller}, destination: frontend/asset-cache[Deployment], dir1:  TCP 8080, dir2: No Connections, diff-type: removed
-source: {ingress-controller}, destination: frontend/webapp[Deployment], dir1:  TCP 8080, dir2: No Connections, diff-type: removed
\ No newline at end of file
+source: {ingress-controller}, destination: frontend/asset-cache[Deployment], dir1:  TCP 8080, dir2: No Connections, diff-type: removed (workload {ingress-controller} removed)
+source: {ingress-controller}, destination: frontend/webapp[Deployment], dir1:  TCP 8080, dir2: No Connections, diff-type: removed (workload {ingress-controller} removed)
```
adisos, line 3: `ingress-controller` is not an added/removed workload.. should be an exception (its a symbolic notation for actual ingress controller pod, and its deployment does not depend if there is or not a Route or Ingress resources).

tests/onlineboutique_workloads_with_ingress/diff_output_from_onlineboutique_workloads.csv
```diff
@@ -1,3 +1,3 @@
 source,destination,dir1,dir2,diff-type
 default/redis-cart[Deployment],default/frontend[Deployment],No Connections,TCP 8080,added
-{ingress-controller},default/frontend[Deployment],No Connections,TCP 8080,added
+{ingress-controller},default/frontend[Deployment],No Connections,TCP 8080,added (workload {ingress-controller} added)
```
adisos, line 3: same here

adisos (changes requested):
added a few initial comments.
for the case of 2 added workloads, the output should be `(workloads A and B added) ` rather than `(workload A added)(workload B added)`

tests/acs-security-demos-no-routes/diff_output_from_acs-security-demos.txt
shireenf-ibm: fixed

shireenf-ibm:
> added a few initial comments. for the case of 2 added workloads, the output should be `(workloads A and B added) ` rather than `(workload A added)(workload B added)`

updated

pkg/netpol/diff/diff.go
```diff
@@ -461,18 +472,32 @@ func getFormatter(format string) (diffFormatter, error) {
...
+	removedDst  bool
+}
+
+// AddedConnsPeers encapsulates an added connection between two peers and indications if any of the peers is new
+type AddedConnsPeers struct {
+	addedConn connlist.Peer2PeerConnection
+	addedSrc  bool
+	addedDst  bool
+}
+
```
adisos, line 488: will it simplify if you remove those new types and just use the type `ConnsPair`,  with additional fields for that type that indicate if it is added/removed/changed, including src/dst info whether it is added/removed?

pkg/netpol/diff/diff_formatter.go
```diff
@@ -54,16 +58,33 @@ func formDiffFieldsDataOfChangedConns(changedConns []*ConnsPair) (netpolsChanged
 	return netpolsRes, ingressRes
 }
 
-func formDiffFieldsDataOfRemovedConns(removedConns []connlist.Peer2PeerConnection) (netpolsRemoved, ingressRemoved []*singleDiffFields) {
+func formDiffFieldsDataOfRemovedConns(removedConns []RemovedConnsPeers) (netpolsRemoved, ingressRemoved []*singleDiffFields) {
```
adisos, line 61: can avoid code duplication for `formDiffFieldsDataOfRemovedConns` and `formDiffFieldsDataOfAddedConns` ?

pkg/netpol/diff/diff.go
```diff
@@ -107,6 +107,8 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
 		da.errors = append(da.errors, newConnectionsAnalyzingError(err))
 		return nil, err
 	}
+	// peers names found in the dirs (some peers may not appear in the conns, so we need to list all peers in the resources)
+	workloadsNames1, workloadsNames2 := caAnalyzer.GetPeersNamesFromDirPath(dirPath1), caAnalyzer.GetPeersNamesFromDirPath(dirPath2)
```
adisos, line 111: can  avoid the changes in `connlist` and just compute the workloads here, on the result returned from `ConnlistFromDirPath`. The function `getIPblocksFromConnList` does something similar - it retrieves the ip-blocks from the connlist result. You can add another function to retrieve the workloads from that result.

pkg/netpol/diff/diff.go
```diff
@@ -394,9 +399,15 @@ func diffConnectionsLists(conns1, conns2 []connlist.Peer2PeerConnection) (Connec
...
+			// removed conn means both Src and Dst exist in peers1, just check if they are not in peers2 too, ignore ips
+			res.removedConns = append(res.removedConns, RemovedConnsPeers{d.firstConn,
+				!(d.firstConn.Src().IsPeerIPType() || eval.IsFakePeer(d.firstConn.Src())) && !peers2[d.firstConn.Src().String()],
+				!(d.firstConn.Dst().IsPeerIPType() || eval.IsFakePeer(d.firstConn.Dst())) && !peers2[d.firstConn.Dst().String()]})
 		case d.secondConn != nil:
-			res.addedConns = append(res.addedConns, d.secondConn)
+			// added conns means Src and Dst are in peers2, check if they didn't exist in peers1 too, , ignore ips/ingress-controller pod
+			res.addedConns = append(res.addedConns, AddedConnsPeers{d.secondConn,
+				!(d.secondConn.Src().IsPeerIPType() || eval.IsFakePeer(d.secondConn.Src())) && !peers1[d.secondConn.Src().String()],
+				!(d.secondConn.Dst().IsPeerIPType() || eval.IsFakePeer(d.secondConn.Dst())) && !peers1[d.secondConn.Dst().String()]})
```
adisos, line 410: can this code be simplified?

pkg/netpol/connlist/connlist.go
```diff
@@ -491,3 +503,8 @@ func (ca *ConnlistAnalyzer) warnBlockedIngress(peerStr string, ingressObjs map[s
 	ca.errors = append(ca.errors, newIngressAnalyzerConnsBlockedWarning(errors.New(warningMsg)))
 	ca.logger.Warnf(warningMsg)
 }
+
+// GetPeersNamesFromDirPath given a dirPath , returns set of names of the peers that were captured in the directory
+func (ca *ConnlistAnalyzer) GetPeersNamesFromDirPath(dirPath string) map[string]bool {
+	return ca.dirToPeersNamesSet[dirPath]
+}
```
adisos, line 510: I think this should not be in a separate function (it assumes that the function `ConnlistFromDirPath` was called before).
Let's changed the returned values of `ConnlistFromDirPath` to include also the list of workloard peers .

pkg/netpol/diff/diff.go
shireenf-ibm: regarding using the `ConnsPair `type for the 3 diff types:
-> in commit [e6ceb4db](https://github.com/np-guard/netpol-analyzer/pull/202/commits/e6ceb4db15ea5d56a06d514bccb2678de433ce6c):
I chose to write the `connectivityDiff struct` , this way :
```
type connectivityDiff struct {
	diffConns []*ConnsPair
```
since all 3 diff types (changed, removed and added) are represented with `[]*ConnsPair ` , and single `ConnsPair ` object contains also the diffType.
however, in order to implement the `ConnectivityDiff interface`, I had to add the func : 
`func (c *connectivityDiff) diffConnectionsByType(diffType string) []*ConnsPair`

-> But, I hesitated if this was efficient or not, so in commit [2fd4cb8f](https://github.com/np-guard/netpol-analyzer/pull/202/commits/2fd4cb8fd3b1832135cf1f4de32dad479e2fe430) 
I chose to return the struct to be as following (the original with the new types for added and removed):
```
type connectivityDiff struct {
	removedConns []*ConnsPair
	addedConns   []*ConnsPair
	changedConns []*ConnsPair
}
```
without any additional funcs to implement the `ConnectivityDiff interface` (and even though the 3 elements of the struct has same type)

pkg/netpol/diff/diff.go
```diff
@@ -98,12 +100,13 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
 		caAnalyzer = connlist.NewConnlistAnalyzer(connlist.WithLogger(da.logger), connlist.WithWalkFn(da.walkFn))
 	}
 	var conns1, conns2 []connlist.Peer2PeerConnection
+	var workloadsNames1, workloadsNames2 []string
 	var err error
-	if conns1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
+	if conns1, workloadsNames1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
 		da.errors = append(da.errors, newConnectionsAnalyzingError(err))
 		return nil, err
 	}
```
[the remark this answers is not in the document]
adisos: also for lines 118-131?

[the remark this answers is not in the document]
shireenf-ibm: no, `err` and `da.errors[i]`  does not have the same type
da.errors is list of `DiffError ` which has (similar to `ConnlistError` an `Error` func `Error() error`.
how ever, the returned `err` is `var err error` which has func `Error() string`
we need to "wrap" the `err` with a struct that implements  the  `DiffError ` (we have same hnadling errors in `Connlist`)

also, `err` does not have an indication of the type of the err (fatal, severe, warning) while `DiffError `  has

pkg/netpol/eval/peer.go
```diff
@@ -17,6 +17,8 @@ type Peer interface {
 	IP() string
 	// IsPeerIPType returns true if  peer is IP address
 	IsPeerIPType() bool
+	// IsFakePeer returns true if the peer was added for analysis goals (not IP neither parsed pod/workload)
+	IsFakePeer() bool
```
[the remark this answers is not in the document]
shireenf-ibm: yes, I can add a func to `eval` package that checks the peer's `pod.Fake` directly (and then will not need these `IsFakePeer`)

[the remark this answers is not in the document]
adisos: let's discuss about additional options here

pkg/netpol/diff/diff.go
```diff
@@ -159,11 +162,23 @@ func getKeyFromP2PConn(c connlist.Peer2PeerConnection) string {
...
-	firstConn  connlist.Peer2PeerConnection
-	secondConn connlist.Peer2PeerConnection
+	firstConn    connlist.Peer2PeerConnection
+	secondConn   connlist.Peer2PeerConnection
+	diffType     string
+	newOrLostSrc bool
+	newOrLostDst bool
 }
 
 // update func of ConnsPair obj, updates the pair with input Peer2PeerConnection, at first or second conn
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/connlist/connlist.go
```diff
@@ -427,10 +431,19 @@ func (ca *ConnlistAnalyzer) getConnectionsBetweenPeers(pe *eval.PolicyEngine) ([
...
+			connsRes = append(connsRes, p2pConnection)
+		}
+		// save the src peerName in the map ca.dirToPeersNamesSet
+		if !srcPeer.IsPeerIPType() {
+			peersSet[srcPeer.String()] = true
 		}
 	}
-	return res, nil
+
+	peersRes := make([]string, 0)
```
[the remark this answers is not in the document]

pkg/netpol/diff/diff.go
```diff
@@ -98,12 +100,13 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
 		caAnalyzer = connlist.NewConnlistAnalyzer(connlist.WithLogger(da.logger), connlist.WithWalkFn(da.walkFn))
 	}
 	var conns1, conns2 []connlist.Peer2PeerConnection
+	var workloadsNames1, workloadsNames2 []string
 	var err error
-	if conns1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
+	if conns1, workloadsNames1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/peer.go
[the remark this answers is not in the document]
shireenf-ibm: done (opened issue #203)

pkg/netpol/connlist/connlist.go
```diff
@@ -409,7 +409,7 @@ func (ca *ConnlistAnalyzer) getConnectionsBetweenPeers(pe *eval.PolicyEngine) ([
 	}
 
 	connsRes := make([]Peer2PeerConnection, 0)
-	peersSet := make(map[string]bool)
+	peers := make([]eval.Peer, 0)
```
adisos, line 412: why not returning `peerList` from line 405?

pkg/netpol/diff/diff.go
```diff
@@ -135,6 +137,22 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
...
+	// list of unique names
+	peersRes := make([]string, len(peersSet))
+	i := 0
+	for peerName := range peersSet {
+		peersRes[i] = peerName
+		i++
+	}
+	return peersRes
+}
+
```
adisos: why not returning `map[string]bool`? (return `peersSet` instead of `peersRes`)?

pkg/netpol/diff/diff.go
```diff
@@ -212,10 +237,10 @@ func (c *ConnsPair) updateNewOrLostFields(isFirst bool, peersList []string) {
 		src, dst = c.secondConn.Src(), c.secondConn.Dst()
 	}
 	// update src/dst status based on the peerList , ignore ips/ingress-controller pod
-	if !(src.IsPeerIPType() || src.IsFakePeer()) && !slices.Contains(peersList, src.String()) {
+	if !(src.IsPeerIPType() || isIngressControllerPeer(src)) && !slices.Contains(peersList, src.String()) {
```
adisos, line 240: here we want to avoid using `slices.Contains`, and just look for `src.String` in a map (`map[string]bool`)

pkg/netpol/connlist/connlist.go
shireenf-ibm: right

pkg/netpol/diff/diff.go
shireenf-ibm: changed

pkg/netpol/diff/diff.go
shireenf-ibm: done

pkg/netpol/diff/diff.go
```diff
@@ -203,6 +221,13 @@ func (c *ConnsPair) isSrcOrDstPeerIPType(checkSrc bool) bool {
 	return (checkSrc && src.IsPeerIPType()) || (!checkSrc && dst.IsPeerIPType())
 }
 
+// helpers to check if a peer is ingress-controller (a peer created while ingress analysis)
+const ingressPodName = "ingress-controller"
+
+func isIngressControllerPeer(peer eval.Peer) bool {
+	return strings.Contains(peer.String(), ingressPodName) // peer.String() == "{"+ingressPodName+"}"
+}
+
```
adisos, line 220: why not comparing to a more specific string `{ingress-controller}`?

shireenf-ibm: changed

pkg/netpol/diff/diff.go
```diff
@@ -215,7 +214,7 @@ func (c *ConnsPair) isSrcOrDstPeerIPType(checkSrc bool) bool {
 const ingressPodName = "ingress-controller"
 
 func isIngressControllerPeer(peer eval.Peer) bool {
-	return strings.Contains(peer.String(), ingressPodName) // peer.String() == "{"+ingressPodName+"}"
+	return peer.String() == "{"+ingressPodName+"}"
```
adisos: can change as follows:

```
const ingressPodName = "{ingress-controller}"

func isIngressControllerPeer(peer eval.Peer) bool {
	return peer.String() == ingressPodName
```

adisos: also, rename `ingressPodName` to `ingressControllerPodName`

shireenf-ibm: done

shireenf-ibm:
hi @adisos , the tests are failing with weird (not really existing) error.
locally, all tests (including lint) successfully pass
how should this be resolved? (before merging into `semantic_diff_cmd` branch)

adisos:
why does the ci fail? 
other than that, no other comments.

Issue #203, opened 2023-08-06
avoid duplicated const in several packages


shireenf-ibm:
currently we have same `const IngressPodName = "ingress-controller"`
in both `ingress-analyzer` (internal to `Connlist`) and `diff` packages 
- to avoid this duplication by exporting  (TBD: where to export it)

adisos:
additional duplication to consider: interfaces `ConnListError` and `DiffEror` are the same

shireenf-ibm:
- [ ] resolve `const IngressPodName = "ingress-controller"` duplication
- [ ]  avoid hard-coded/duplicated error messages for testing. Ideally, we should have a pkg with defined error strings as const values , that will be used both for error generation and for testing errors strings cross-packages. (see [example](https://github.com/np-guard/netpol-analyzer/pull/244#discussion_r1345735863))

PR #218, opened 2023-08-15
issue 215 - fix errors appending bug

shireenf-ibm:
the problem was : `connlistAnalyzer.Errors() `  after first call to `caAnalyzer.ConnlistFromDirPath(dirPath1)`, were appended twice into diffAnalyzer Errors list (once after each call to the ConnlistAnalyzer)
fix: errors returned by first call, should be appended only if we are returning from the `calling func`.
After finishing with all calls to the `connlistAnalyzer` , all `connlistAnalyzer.Errors()` are appended to the diffAnalyzer Errors list 

** also, added the example folder with malformed yamls
** I suggest adding a `--fail` flag to enable running `WithStopOnError` from cmd-line too (helps with local testing)
if it is acceptable, should be added with this PR or new one?

### DIFF
--- a/pkg/netpol/diff/diff.go
+++ b/pkg/netpol/diff/diff.go
@@ -87,10 +87,13 @@ func (da *DiffAnalyzer) Errors() []DiffError {
 	return da.errors
 }
 
-// return true if has fatal error or severe with flag stopOnError
-func (da *DiffAnalyzer) stopProcessing() bool {
-	for idx := range da.errors {
-		if da.errors[idx].IsFatal() || da.stopOnError && da.errors[idx].IsSevere() {
+// loops the errors that were returned from the connlistAnalyzer
+// (as only connlistAnalyzer.Errors() may contain severe errors; all other DiffAnalyzer errors are fatal),
+// returns true if has severe error with flag stopOnError
+// if it has fatal error , means it returned before calling this,
+func (da *DiffAnalyzer) stopProcessing(caErrors []connlist.ConnlistError) bool {
+	for _, e := range caErrors {
+		if da.stopOnError && e.IsSevere() {
 			return true
 		}
 	}
@@ -119,19 +122,22 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
 	var workloadsNames1, workloadsNames2 map[string]bool
 	var err error
 	if conns1, workloads1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
-		da.errors = append(da.errors, newConnectionsAnalyzingError(err, true, false))
+		// append all fatal/severe errors and warnings returned by caAnalyzer then return because of the fatal err
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
 		return nil, err
 	}
-	da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
-	if da.stopProcessing() {
+	if da.stopProcessing(caAnalyzer.Errors()) {
+		// if true, before returning, append the caAnalyzer.Errors to DiffAnalyzer.Errors() to be exported
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
 		return &connectivityDiff{}, nil
 	}
 	if conns2, workloads2, err = caAnalyzer.ConnlistFromDirPath(dirPath2); err != nil {
-		da.errors = append(da.errors, newConnectionsAnalyzingError(err, true, false))
+		// append all fatal/severe errors and warnings returned by caAnalyzer then return because of the fatal err
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
 		return nil, err
 	}
-	da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
-	if da.stopProcessing() {
+	da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors()) // append all caAnalyzer.Errors() as we finished calling its funcs
+	if da.stopProcessing(caAnalyzer.Errors()) {              // checks if the second dirPath raised any severe errors
 		return &connectivityDiff{}, nil
 	}
 	workloadsNames1, workloadsNames2 = getPeersNamesFromPeersList(workloads1), getPeersNamesFromPeersList(workloads2)
--- a/pkg/netpol/diff/diff_test.go
+++ b/pkg/netpol/diff/diff_test.go
@@ -152,9 +152,9 @@ type testErrEntry struct {
 	isCaWarning     bool
 	isFormattingErr bool
 	format          string
+	emptyRes        bool
 }
 
-var caErrType = &connectionsAnalyzingError{}     // error returned from a func on the ConnlistAnalyzer object
 var formattingErrType = &resultFormattingError{} // error returned from getting/writing output format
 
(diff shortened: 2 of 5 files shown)


pkg/netpol/diff/diff.go
```diff
@@ -119,19 +122,22 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
...
 	var err error
 	if conns1, workloads1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
-		da.errors = append(da.errors, newConnectionsAnalyzingError(err, true, false))
+		// append all fatal/severe errors and warnings returned by caAnalyzer then return because of the fatal err
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
 		return nil, err
 	}
-	da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
-	if da.stopProcessing() {
+	if da.stopProcessing(caAnalyzer.Errors()) {
```
adisos, line 129: why isn't `da.errors` sent to this function?

pkg/netpol/diff/diff.go
```diff
@@ -119,19 +122,22 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
...
+		// if true, before returning, append the caAnalyzer.Errors to DiffAnalyzer.Errors() to be exported
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
```
adisos, line 131: why need to append here after appended in line 126?

adisos (changes requested):
Added a few comments.
Please open an issue for adding a --fail flag (and tests), and handle this in a separate branch/PR.

pkg/netpol/diff/diff.go
```diff
@@ -87,10 +87,13 @@ func (da *DiffAnalyzer) Errors() []DiffError {
...
-func (da *DiffAnalyzer) stopProcessing() bool {
-	for idx := range da.errors {
-		if da.errors[idx].IsFatal() || da.stopOnError && da.errors[idx].IsSevere() {
+// loops the errors that were returned from the connlistAnalyzer
+// (as only connlistAnalyzer.Errors() may contain severe errors; all other DiffAnalyzer errors are fatal),
+// returns true if has severe error with flag stopOnError
+// if it has fatal error , means it returned before calling this,
+func (da *DiffAnalyzer) stopProcessing(caErrors []connlist.ConnlistError) bool {
+	for _, e := range caErrors {
+		if da.stopOnError && e.IsSevere() {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/diff/diff.go
shireenf-ibm: - `stopProcessing()` is called only after returning from a Connlist Analyzer func , so actually we need to check if the connlist Errors list include severe errors and if the `da.stopOnError` flag is on .
- only Connlist Errors may include severe error/ warning beside to fatal (all of the errors raised in diff package / by the Diff analyzer are fatal , and when raised cause to stop running regardless the `stopOnError `flag
- we need to call `stopProcessing() ` after each `caAnalyzer.ConnlistFromDirPath()`, so if there is a severe error and the `stopOnError` flag is on, we need to stop, but we can not append the connlist errors to the diff analyzer errors before calling it, otherwise all connlist errors that were returned from the first `caAnalyzer.ConnlistFromDirPath(dirPath1)` call , will be appended again when returning from the `caAnalyzer.ConnlistFromDirPath(dirPath2)` (if the stopOnError flag was off / if there were only warnings -  that what caused the bug initially)

shireenf-ibm:
[commit  avoid duplicated error interfaces](https://github.com/np-guard/netpol-analyzer/pull/218/commits/3b74518857f26b687e3f02f82fecd0dea5fd469a) resolves [following comment: ](https://github.com/np-guard/netpol-analyzer/issues/203#issuecomment-1676269910) in issue #203

kept the original interface names for each pkg as they are exported that way

shireenf-ibm:
[commit fixing diff tests checks](https://github.com/np-guard/netpol-analyzer/pull/218/commits/3376027320b72626ed1df91015752c7c8c8a514d)  improves testing checks.

also, splet the "old" (previous version of `func TestDiffErrors`)  to 3 different funcs; as it included many if conditions

pkg/netpol/diff/diff.go
```diff
@@ -122,19 +124,22 @@ func (da *DiffAnalyzer) ConnDiffFromDirPaths(dirPath1, dirPath2 string) (Connect
 	var workloadsNames1, workloadsNames2 map[string]bool
 	var err error
 	if conns1, workloads1, err = caAnalyzer.ConnlistFromDirPath(dirPath1); err != nil {
-		da.errors = append(da.errors, newConnectionsAnalyzingError(err, true, false))
+		// append all fatal/severe errors and warnings returned by caAnalyzer then return because of the fatal err
+		da.appendConnlistErrorsToDiffErrors(caAnalyzer.Errors())
```
adisos, line 128: lines 126-144 seems to have code duplication -- can you move common code to a separate smaller function?

pkg/netpol/diff/diff_test.go
adisos, line 1: Separation to functions of different error levels helps.
What is the relevant test that had the issue of multiple (duplicated) error instances before the fix in this branch? what was its number of errors before and after the fix?

pkg/netpol/diff/diff.go
shireenf-ibm: done

pkg/netpol/diff/diff_test.go
shireenf-ibm: before this fix, all connlist analyzer's errors (`caAnalyzer.Errors()`) after analyzing dir1 were appended to the diffError list twice in case the diff analysis **didn't stop after dir1 analysis** ( i.e. after returning from  `caAnalyzer.ConnlistFromDirPath(dirPath1)`.).

(for all tests , especially tests with severe errors running without stopOnError )

for example :
for running following command :
` ./bin/k8snetpolicy diff --dir1 tests/bad_yamls/not_a_k8s_resource.yaml --dir2 tests/ipblockstest/`
after the fix, eventually the diffErrors len is 2 (as `dir1` has two errors (a warning and a severe error) while dir2 has no errors at all )
before the fix, **the length of diffErrors was 4** (the errors from dir1 were duplicated)

pkg/netpol/diff/diff_test.go
```diff
@@ -172,171 +172,204 @@ func TestDiff(t *testing.T) {
...
+			// dirty directory, includes 3 severe errors
+			// when running without stopOnError we expect to see 6 severe errors (3 for each dir flag)
+			// but when running with stopOnError we expect to see only 1 , and then stops
+			name:                           "both dirs return severe errors on their malformed yaml files",
+			dir1:                           "dirty",
+			dir2:                           "dirty",
+			firstErrStr:                    "YAML document is malformed",
+			expectedErrNumWithoutStopOnErr: 6,
+			expectedErrNumWithStopOnErr:    1,
 		},
```
shireenf-ibm, line 315: this is another example,
after the fix we see `expectedErrNumWithoutStopOnErr: 6,`  (as `dirty` dir has 3 errors , so for dir1+dir2 it is 6)
before the fix we would get  `expectedErrNumWithoutStopOnErr: 9,`

adisos (approved):

PR #227, opened 2023-08-27
DOT format for diff

shireenf-ibm:
#154


pkg/netpol/diff/diff_formatter.go
```diff
@@ -215,3 +215,130 @@ func (cs *diffFormatCSV) writeDiffOutput(connsDiff ConnectivityDiff) (string, er
 func (cs *diffFormatCSV) singleDiffLine(d *singleDiffFields) string {
 	return fmt.Sprintf("%s;%s;%s;%s;%s;%s", d.diffType, d.src, d.dst, d.dir1Conn, d.dir2Conn, d.workloadDiffInfo)
 }
+
+// /////////////////////////
+// diffFormatDOT: implements the diffFormatter interface for dot output format
```
adisos, line 220: split every fomatter type to a separate file

pkg/netpol/diff/diff_formatter.go
```diff
@@ -215,3 +215,130 @@ func (cs *diffFormatCSV) writeDiffOutput(connsDiff ConnectivityDiff) (string, er
...
+type diffFormatDOT struct {
+}
+
+const (
+	dotHeader  = "digraph {"
+	dotClosing = "}"
+)
```
adisos, line 227: we have these const value also defined in connlist pkg, consider moving this to common pkg?

pkg/netpol/diff/diff_formatter.go
```diff
@@ -215,3 +215,130 @@ func (cs *diffFormatCSV) writeDiffOutput(connsDiff ConnectivityDiff) (string, er
...
+	// write graph
+	allLines := []string{dotHeader}
+	allLines = append(allLines, peersLines...)
+	allLines = append(allLines, edgeLines...)
+	allLines = append(allLines, ingressAnalyzerEdges...)
+	allLines = append(allLines, dotClosing)
+	return strings.Join(allLines, newLine), nil
+}
+
+func getEdgesAndPeersLinesByCategory(connsPairs []*ConnsPair, peersSet map[string]bool) (pLines, cEdges, iEdges []string) {
```
adisos, line 267: rename `pLines, cEdges, iEdges` to more meaningful names.
you can use the actual names of the returned objects `peersLines, connsEdges, ingressEdges` , and then change so that they are not redefined in the function body.

pkg/netpol/diff/diff_formatter.go
```diff
@@ -215,3 +215,130 @@ func (cs *diffFormatCSV) writeDiffOutput(connsDiff ConnectivityDiff) (string, er
...
+	peersLines := make([]string, 0)
+	connsEdges := make([]string, 0)
+	ingressEdges := make([]string, 0)
+	for _, connsPair := range connsPairs {
+		src, dst, isIngress := getConnPeersStrings(connsPair)
+		// add peers lines (which still not in the set)
```
adisos, line 273:
```suggestion
		// add peers lines (which are still not in the set)
```

pkg/netpol/diff/diff_formatter.go
```diff
@@ -215,3 +215,130 @@ func (cs *diffFormatCSV) writeDiffOutput(connsDiff ConnectivityDiff) (string, er
...
+)
+
+func addEdgesLines(connsPair *ConnsPair) string {
+	src, dst, _ := getConnPeersStrings(connsPair)
+	firstConn, secondConn := getDirsConnsStrings(connsPair)
+	switch connsPair.diffType {
+	case nonChangedType:
+		return getEdgeLine(src, dst, firstConn, nonChangedConnColor)
+	case changedType:
+		return getEdgeLine(src, dst, firstConn+"->"+secondConn, changedConnColor)
```
adisos, line 327: can we change the format of changed conn edge?
I think `->` might be confusing
one option: `new_conn (was: old_conn)`
do you have other ideas?

adisos (changes requested):
looks good!
Added a few comments

pkg/netpol/diff/diff_formatter.go
adisos: let's replace `was` with `old` instead

shireenf-ibm: done

adisos (commented):
looks good!
please also update documentation/command help -- that this format is supported for the `diff` command.

shireenf-ibm:
> looks good! please also update documentation/command help -- that this format is supported for the `diff` command.

updated in readme.
already included in the description msg of `output` flag for diff (generated auto from `diff.ValidDiffFormats`)

adisos (approved):

Issue #234, opened 2023-09-11
namespace analysis issue


adisos:
The analyzer should add label `kubernetes.io/metadata.name`  to any namespace in the analysis, even if its manifest is given.

PR #240, opened 2023-09-19
namespace analysis 

shireenf-ibm:
#234 
when having a namespace manifest , the `NamespaceFromCoreObject` func is called, if the required "kubernetes.io/metadata.name" label is missing in the manifest, it'll be added


pkg/netpol/eval/internal/k8s/namespace.go
```diff
@@ -23,6 +23,8 @@ type Namespace struct {
 	Labels map[string]string
 }
 
+const K8sNsNameLabelKey = "kubernetes.io/metadata.name"
```
adisos, line 26:
```suggestion
// The Kubernetes API server sets this label on all namespaces
const K8sNsNameLabelKey = "kubernetes.io/metadata.name"
```

pkg/netpol/eval/internal/k8s/namespace.go
```diff
@@ -36,5 +38,10 @@ func NamespaceFromCoreObject(ns *corev1.Namespace) (*Namespace, error) {
 	for k, v := range ns.Labels {
 		n.Labels[k] = v
 	}
+
+	if _, ok := n.Labels[K8sNsNameLabelKey]; !ok {
+		n.Labels[K8sNsNameLabelKey] = ns.Name
+	}
```
adisos, line 44:
```suggestion
	// if missing, the label set by k8s API server must be added to the namespace labels
	if _, ok := n.Labels[K8sNsNameLabelKey]; !ok {
		n.Labels[K8sNsNameLabelKey] = ns.Name
	}
```

adisos (changes requested):
looks good.
please add a relevant test -- policy manifest that uses this label, with a namespace manifest defined and missing this label (so that the analysis would not work correctly before this change).

shireenf-ibm:
in the new test:
the connection from workload-b to workload-a (i.e. `hello-world/workload-b[Deployment] => hello-world/workload-a[Deployment] : TCP 8050`) would not be captured if the `kubernetes.io/metadata.name` not added by the analyzer

Issue #200, opened 2023-08-01
SemanticDiff tests from NCA which are not imported yet


shireenf-ibm:
This issue lists the k8s semanticDiff in NCA which could not be used for netpol-analyzer testing and why not, to consider a future support:

- [x]  [semanticDiff-same-topologies-scheme](https://github.com/IBM/network-config-analyzer/blob/3c485449a02b7f28030a87362e6c78d83a5a1a48/tests/k8s_testcases/example_policies/tests-different-topologies/semanticDiff-same-topologies-scheme.yaml) : 

- [x]  [semanticDiff-different-topologies-scheme.yaml](https://github.com/IBM/network-config-analyzer/blob/ebe0e452759ef3a06cfdcf79cfd4526ae2c413ea/tests/k8s_testcases/example_policies/tests-different-topologies/semanticDiff-different-topologies-scheme.yaml) (queries : semantic_diff_a_to_b, semantic_diff_b_to_a, semantic_diff_disjoint_old1_config_a, semantic_diff_a_to_b_with_ipBlock)

 (**First task below**)
the podList in these tests contains multiple pods with different labels, but same `ownerReferences` name, a case that is not supported in netpol-analyzer 

------------------------------------------------------

- [x]  [namedPorts-scheme.yaml](https://github.com/IBM/network-config-analyzer/blob/3c485449a02b7f28030a87362e6c78d83a5a1a48/tests/k8s_testcases/example_policies/tests-different-topologies/namedPorts-scheme.yaml):

1. First SemanticDiff in the file:  **RESOLVED** :
this test is legal, the netpols select all pods and try to connect only to the pods with specific named-ports. currently, netpol-analyzer, raises an error in case a selected pod does not contain a named port/ that specific named-port (a bug to be resolved in #199)
2. Second SemanticDiff in the file : 
running the `list` command on the relevant dirs, gives different connections for that pod each time >> **same problem as first task below** (see comment below)

-------------------------------------

- [x]  [semanticDiff-multiple-resources-scheme.yaml](https://github.com/IBM/network-config-analyzer/blob/ebe0e452759ef3a06cfdcf79cfd4526ae2c413ea/tests/k8s_testcases/example_policies/tests-different-topologies/semanticDiff-multiple-resources-scheme.yaml):(first query) - **first task below :** policy's rule in this test select the pods with `app: bank-ui-selector`, which is not shared by all pods with same `ownerReferences` name 

[...]

PR #244, opened 2023-09-28
Issue 200 err message and more diff tests based nca

shireenf-ibm:
#200 
#128


pkg/netpol/eval/resources.go
```diff
@@ -203,12 +207,34 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+
 func (pe *PolicyEngine) upsertWorkload(rs interface{}, kind string) error {
 	pods, err := k8s.PodsFromWorkloadObject(rs, kind)
 	if err != nil {
 		return err
 	}
-	for _, podObj := range pods {
+	for index, podObj := range pods {
+		if index < 1 { // check only once, as all replicas has same ownerReference and same labels
+			if err := pe.checkLegalOwnerRefNameToPodLabels(podObj.Owner.Name, podObj.Labels); err != nil {
```
adisos, line 234: Should be checked only from `upsertPod` and not from `upsertWorkload`.
In the hierarchy of template owners there are cases where controllers add labels for the  generated resources, so that a full match is not necessarily the case. For example, `CronJob` creates pods with added labels `controller-uid`. Another example: `Deployment` creates `ReplicaSet` with labels `pod-template-hash`.

pkg/netpol/eval/resources.go
```diff
@@ -22,10 +23,11 @@ type (
...
 	// and allows querying it for allowed or denied connections.
 	PolicyEngine struct {
-		namspacesMap map[string]*k8s.Namespace                // map from ns name to ns object
-		podsMap      map[string]*k8s.Pod                      // map from pod name to pod object
-		netpolsMap   map[string]map[string]*k8s.NetworkPolicy // map from namespace to map from netpol name to its object
-		cache        *evalCache
+		namspacesMap      map[string]*k8s.Namespace                // map from ns name to ns object
+		podsMap           map[string]*k8s.Pod                      // map from pod name to pod object
+		netpolsMap        map[string]map[string]*k8s.NetworkPolicy // map from namespace to map from netpol name to its object
+		ownersToLabelsMap map[string]map[string]string             // map from ownerReference name to its labels map
```
adisos, line 29:
```suggestion
		ownersToLabelsMap map[string]map[string]string             // map from pods' ownerReference name to its labels map
```

adisos: should be in the context of a namespace (there could be two owners of the same name in different namespaces).

adisos (changes requested):

tests/onlineboutique_with_replicas_and_variants/pods.yaml
```diff
@@ -955,7 +955,6 @@ items:
     labels:
       app: loadgenerator
       pod-template-hash: 555fbdc87d
-      unhealthy: "true"
```
adisos, line 958: Please check `eval_test.go` lines 1653-1661. The test specifically specifies "loadgenerator pod has 3 replicas but one with variant on label" , and this is required for the test. Doesn't your change break this test?

shireenf-ibm: my changes, cause the cache hits count to be raised (was 25 now 49, i.e before 2 replicas were counted now 3)
this does not break the test , since the `expectedCacheHits` is not const , it is computed dynamically through the test run (in line 1677) 
so both the actual and expected values are changed and equal in both cases

shireenf-ibm: however since currently the changes affect only connectivity analysis commands and not eval command. (the check of labels per owner ref. is moved to `getpeersList func)` , and since this test is used only in `eval_test.go` , 
**reverted this change**

pkg/netpol/eval/resources.go
```diff
@@ -285,20 +327,25 @@ func (pe *PolicyEngine) HasPodPeers() bool {
 }
 
 // createPodOwnersMap creates map from workload str to workload peer object
```
adisos, line 329:
```suggestion
// createPodOwnersMap creates map from workload str to workload peer object
// returns error if there are two pods of same owner but different set of labels, since cannot map inconsistent pods to a workload
```

adisos (changes requested):

tests/onlineboutique_with_replicas_and_variants/pods.yaml
adisos: Now that this change was reverted it's ok. (before revert, the change did not break the test, but had made it the same as the test above it , and thus the functionality required for the test was no longer tested... there was intentionally one test that has a pod with 3 replicas, and another test that has pod with 3 replicas but one with variant on the labels).

pkg/netpol/eval/resources.go
```diff
@@ -203,6 +209,41 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+	}
+	ownerLabels, ok := pe.ownersToLabelsMap[namespace][ownerRefName]
+	if !ok { // add the new ownerReference with pod labels to the map
+		pe.ownersToLabelsMap[namespace][ownerRefName] = podLabels
+		return nil
+	}
+	// compare ownerLabels with podLabels
+	if !reflect.DeepEqual(ownerLabels, podLabels) {
+		return errors.New("Resources not supported for connectivity analysis. Pods with the ownerReferences' Name: " + ownerRefName +
+			" have different labels. Some labels' keys with different values: " + strings.Join(diffBetweenLabels(podLabels, ownerLabels), ","))
```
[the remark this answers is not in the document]
shireenf-ibm: In order to print these messages I need also to save the first pod name which its labels are saved in the new map as the ownerRef's labels.
Instead of using a new struct with pod name and its labels.
I changed the map to be from namespace to pods' ownerRef. name to its first pod.
then when comparing, I can compare curr. pod labels with first pod's labels and raise the relevant error message when we meet the first different pod

pkg/netpol/eval/resources.go
```diff
@@ -203,6 +209,41 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+	if _, ok := pe.ownersToLabelsMap[namespace]; !ok { // add the namespace to the map
+		pe.ownersToLabelsMap[namespace] = make(map[string]map[string]string)
+	}
+	ownerLabels, ok := pe.ownersToLabelsMap[namespace][ownerRefName]
+	if !ok { // add the new ownerReference with pod labels to the map
+		pe.ownersToLabelsMap[namespace][ownerRefName] = podLabels
+		return nil
+	}
+	// compare ownerLabels with podLabels
+	if !reflect.DeepEqual(ownerLabels, podLabels) {
```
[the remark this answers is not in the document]

pkg/netpol/eval/resources.go
```diff
@@ -203,6 +207,54 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
 	return nil
 }
 
+// checkLegalOwnerRefNameToPodLabels returns error if there are pod resources with same ownerReferences name but different labels
+func (pe *PolicyEngine) checkLegalOwnerRefNameToPodLabels(newPod *k8s.Pod) error {
```
adisos, line 211: rename function to be more specific/relevant, such as `checkConsistentLabelsForPodsOfSameOwner`

pkg/netpol/eval/resources.go
```diff
@@ -256,6 +308,7 @@ func (pe *PolicyEngine) deletePod(p *corev1.Pod) error {
 	if podObj, ok := pe.podsMap[podName]; ok {
 		// delete relevant workload entries from cache if all pods per owner are deleted
 		pe.cache.deletePod(podObj, podName)
+		delete(pe.ownersToLabelsMap, podObj.Owner.Name)
```
adisos, line 311: if there are remaining pods of this owner, then the representative pod should be replaced rather than deleted.

pkg/netpol/eval/resources.go
```diff
@@ -203,6 +207,54 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+	if !ok { // add the new ownerReference with this new pod
+		pe.ownersToLabelsMap[newPod.Namespace][newPod.Owner.Name] = newPod
+		return nil
+	}
+	// compare the owner first pod's labels with new pod's Labels
+	return diffBetweenPodsLabels(firstPod, newPod)
+}
+
+// helper: given two pods of same owner, returns an error if there are diffs between the pods' labels maps
+func diffBetweenPodsLabels(firstPod, newPod *k8s.Pod) error {
```
adisos, line 228: better split to a function that handles the error messages and a function that computes the labels gap details.

adisos (changes requested):

pkg/netpol/connlist/connlist_test.go
```diff
@@ -382,3 +458,12 @@ func TestWithFocusWorkloadWithReplicasConnections(t *testing.T) {
...
 	require.NotContains(t, out, "kube-system/calico-node[DaemonSet] => kube-system/calico-node[DaemonSet] : All Connections")
 }
+
+func TestConnlistAnalyzerIllegalPodList(t *testing.T) {
+	dirPath := filepath.Join(testutils.GetTestsDir(), "semanticDiff-same-topologies-illegal-podlist")
+	analyzer := NewConnlistAnalyzer(WithIncludeJSONManifests())
+	res, _, err := analyzer.ConnlistFromDirPath(dirPath)
+	require.Nil(t, res)
+	require.Contains(t, err.Error(), "Input Pod resources are not supported for connectivity analysis."+
+		" Found Pods of the same owner but different set of labels.")
```
adisos, line 468: Current error message output:
```
$ ./bin/k8snetpolicy list --dirpath tests/semanticDiff-same-topologies-illegal-podlist/ --include-json
Error: Input Pod resources are not supported for connectivity analysis. Found Pods of the same owner but different set of labels. Pod demo/account-query has label app with value account-query-selector, and Pod demo/bank-ui has label app with value bank-ui-selector

```
few more issues to address:
- Missing `.` at the end of the last sentence.
- Replace `has label (key) with value (val)` with `has label: key=val`
- Add details about the pods owner `Found pods of the same owner (namespace/owner-name)`

adisos (changes requested):

pkg/netpol/eval/resources.go
shireenf-ibm: - currently I raise the error with first diff found, 
is replacing these lines of computing the error msg and raising it (e.g. line 240-241) by calling a def that handle error msg writing (e.g return computeErrMsg(...) is enough , 
or   do you prefer to return as result all differences found and after that write the error message with any?

adisos: it is sufficient to return the first difference found

pkg/netpol/connlist/connlist_test.go
```diff
@@ -465,5 +465,5 @@ func TestConnlistAnalyzerIllegalPodList(t *testing.T) {
 	res, _, err := analyzer.ConnlistFromDirPath(dirPath)
 	require.Nil(t, res)
 	require.Contains(t, err.Error(), "Input Pod resources are not supported for connectivity analysis."+
-		" Found Pods of the same owner but different set of labels.")
+		" Found Pods of the same owner demo/cog-agents but with different set of labels.")
```
adisos, line 468: I think we should avoid hard-coded/duplicated error messages for testing. Ideally, we should have a pkg with defined error strings as const values , that will be used both for error generation and for testing errors strings cross-packages. You can open another issue so we consider handling this later.

pkg/netpol/eval/resources.go
```diff
@@ -207,52 +207,67 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+	if key, firstVal, newVal := diffBetweenPodsLabels(firstPod, newPod); key != "" {
+		return writeLabelsDifferencesError(firstPod, newPod, key, firstVal, newVal) // err
+	}
+	return nil
 }
 
-// helper: given two pods of same owner, returns an error if there are diffs between the pods' labels maps
-func diffBetweenPodsLabels(firstPod, newPod *k8s.Pod) error {
+// helper: writeLabelsDifferencesError computes the error message of the pod's different labels
+func writeLabelsDifferencesError(firstPod, newPod *k8s.Pod, key, firstVal, newVal string) error {
```
adisos, line 231: The function name is misleading, as it does not write anything. Consider changing to `generateLabelsDiffError`

pkg/netpol/eval/resources.go
```diff
@@ -207,52 +207,67 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
-	return diffBetweenPodsLabels(firstPod, newPod)
+	if key, firstVal, newVal := diffBetweenPodsLabels(firstPod, newPod); key != "" {
+		return writeLabelsDifferencesError(firstPod, newPod, key, firstVal, newVal) // err
+	}
+	return nil
 }
 
-// helper: given two pods of same owner, returns an error if there are diffs between the pods' labels maps
-func diffBetweenPodsLabels(firstPod, newPod *k8s.Pod) error {
+// helper: writeLabelsDifferencesError computes the error message of the pod's different labels
```
adisos, line 230: change comment: "error message of the gap between two pods' labels "

pkg/netpol/eval/resources.go
```diff
@@ -207,52 +207,67 @@ func (pe *PolicyEngine) upsertNamespace(ns *corev1.Namespace) error {
...
+	case newVal == "":
+		errMsgPart2 = fmt.Sprintf(keyMissingErr, firstPodStr, key, firstVal, newPodStr, key)
+	default: // both values are not empty
+		errMsgPart2 = fmt.Sprintf(differentValuesErr, newPodStr, key, newVal, firstPodStr, key, firstVal)
+	}
+	return errors.New(errMsgPart1 + errMsgPart2)
+}
 
+// helper: given two pods of same owner, if there are diffs between the pods' labels maps returns first captured diff components,
+// i.e. the different label's key and the different values / empty val if the key is missing in one pod's labels
```
adisos, line 253: add to function description: if there is no diff, returns empty key (with empty values)

adisos (changes requested):
looks good, just a few more comments.

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: added as a new task in #203

adisos (approved):

PR #249, opened 2023-10-12
simplifying dir_not_exist error

shireenf-ibm:
since this will be deprecated when refactoring
I thought it is efficient enough to return a description of the error instead of the original error returned by `walkFunc `


pkg/netpol/scan/error_types.go
```diff
@@ -98,7 +100,11 @@ func (err *FailedReadingFileError) Unwrap() error {
 }
 
 func (err *FailedAccessingDirError) Error() string {
-	return fmt.Sprintf("error accessing directory: %v", err.origErr)
+	if errors.Is(err.origErr, os.ErrNotExist) {
+		return fmt.Sprintf("directory: <%s> was not found", err.dirPath)
```
adisos, line 104:
```suggestion
		return fmt.Sprintf("directory %s was not found", err.dirPath)
```

pkg/netpol/scan/error_types.go
```diff
@@ -98,7 +100,11 @@ func (err *FailedReadingFileError) Unwrap() error {
...
+	}
+
+	return fmt.Sprintf("failed reading contents of directory: <%s> ", err.dirPath)
```
adisos, line 107:
```suggestion
	return fmt.Sprintf("failed reading contents of directory %s ", err.dirPath)
```

adisos (commented):
Looks good,  just small comments.

adisos (approved):

Issue #110, opened 2023-03-28
tests improvements


adisos:
- [ ] use assertion package consistently for all unit tests
- [ ] convert to table-driven tests

adisos:
table-driven tests:
https://adlerhsieh.com/blog/a-deep-dive-into-table-driven-testing-in-golang
https://www.gopherguides.com/articles/table-driven-testing-in-parallel 
https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721
https://medium.com/code-chasm/unit-test-in-go-8d17d3910818

PR #250, opened 2023-10-18
Issue 110 tests improvements connlist pkg

shireenf-ibm:
issue #110 

- this PR , includes tests improvements to `connlist_test.go` file
all test `funcs` run in parallel, and for each `test func, all cases run in parallel too.`

- tried to split `TestXXX` funcs by "testing cases" to avoid `if else` conditions in the test functionality

- this PR does not handle generating/overriding expected output files (prefer to keep it for a separate PR, want to study if it is possible to write a workflow that could be run by `go test --override` (for example))


pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,469 +1,559 @@
...
-			}
-		}
+	for _, tt := range cases {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			analyzerOpts := []ConnlistAnalyzerOption{WithIncludeJSONManifests()}
+			_, res, err := getConnlistFromDirPathRes(analyzerOpts, tt.dirName)
+			require.Empty(t, res, "test: %q", tt.name)
+			require.Contains(t, err.Error(), tt.errorStrContains, "test: %q", tt.name)
```
adisos, line 394: If there is mismatch in err str, provide a more detailed message, including also info about the actual err str vs the expected one, in addition to the test name.

adisos (commented):
Looks very good!
Added a few comments.

adisos:
> issue #110
> 
>     * this PR , includes tests improvements to `connlist_test.go` file
>       all test `funcs` run in parallel, and for each `test func, all cases run in parallel too.`
> 
>     * tried to split `TestXXX` funcs by "testing cases" to avoid `if else` conditions in the test functionality
> 
>     * this PR does not handle generating/overriding expected output files (prefer to keep it for a separate PR, want to study if it is possible to write a workflow that could be run by `go test --override` (for example))

As for generating/overriding expected output files  -- please consider we also need to be able to run this locally, not necessarily in a workflow.

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,469 +1,559 @@
...
+// helping func - creates ConnlistAnalyzer with desired opts and returns the analyzer with connlist from provided directory
+func getConnlistFromDirPathRes(opts []ConnlistAnalyzerOption, dirName string) (*ConnlistAnalyzer, []Peer2PeerConnection, error) {
+	analyzer := NewConnlistAnalyzer(opts...)
+	res, _, err := analyzer.ConnlistFromDirPath(getDirPathFromDirName(dirName))
+	return analyzer, res, err
 }
 
-const expectedOutputFileNamePrefix = "connlist_output."
+// helping func - creates the analyzer , gets connlist and writes it to string and verifies results
+func verifyConnlistAnalyzeOutputVsExpectedOutput(t *testing.T, analyzerOptions []ConnlistAnalyzerOption, dirName,
```
adisos, line 33: Since this function can be called from a loop that runs a single test dir with various output formats, the message of a failed test should also include the output format  it was run with (in addition to the test name) , so that one can debug a failure more easily.

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,469 +1,559 @@
...
+}
 
 // TestConnList tests the output of ConnlistFromDirPath() for valid input resources
 func TestConnList(t *testing.T) {
-	testingEntries := []testEntry{
+	t.Parallel()
+	cases := []struct {
+		name          string
+		testDirName   string
+		outputFormats []string
```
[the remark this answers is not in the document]
shireenf-ibm: instead , I have represented the test name by the dir name for `TestConnList`  and added the format to the "debug" msg as suggested.
and "computed" the test name from both `dir name` and `focus_workload name` for `TestConnListWithFocusWorkload`
[commit ](https://github.com/np-guard/netpol-analyzer/pull/250/commits/e38a51d4daa5f553da05ab283cc1b7335ed0632c)

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: [commit](https://github.com/np-guard/netpol-analyzer/pull/250/commits/7fd2afb3a9c27ba5ba0955ffc13bfc7dfc79c733)

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,469 +1,559 @@
...
+func verifyConnlistAnalyzeOutputVsExpectedOutput(t *testing.T, analyzerOptions []ConnlistAnalyzerOption, dirName,
+	expectedOutputFileName, testName string) {
+	analyzer, res, err := getConnlistFromDirPathRes(analyzerOptions, dirName)
+	require.Nil(t, err, "test: %q", testName)
+	output, err := analyzer.ConnectionsListToString(res)
+	require.Nil(t, err, "test: %q", testName)
+	expectedOutputFile := filepath.Join(getDirPathFromDirName(dirName), expectedOutputFileName)
+	expectedOutput, err := os.ReadFile(expectedOutputFile)
+	require.Nil(t, err, "test: %q", testName)
+	require.Equal(t, string(expectedOutput), output, "test: %q", testName)
```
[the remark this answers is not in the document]
shireenf-ibm: [commit](https://github.com/np-guard/netpol-analyzer/pull/250/commits/dadec5053c8976b40ef4ad6d574d59b8ee1fe8d7)

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,44 +1,105 @@
...
+		actualOutputFile := filepath.Join(getDirPathFromDirName(dirName), actualOutputFileName)
+		fp, err := os.Create(actualOutputFile)
+		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
+		_, err = fp.WriteString(actualOutput)
+		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
+		t.Fatalf("output mismatch for %s, actual output file %q vs expected output file: %q", getDebugMsgWithTestNameAndFormat(testName, format),
+			actualOutputFile, expectedOutputFile)
+	}
+	require.Equal(t, string(expectedOutput), actualOutput, getDebugMsgWithTestNameAndFormat(testName, format))
+}
```
adisos, line 66:
```suggestion
func checkActualVsExpectedOutputMatch(t *testing.T, testName, dirName, expectedOutputFileName, actualOutput, format string) {
	actualOutputFileName := "actual_" + expectedOutputFileName
	// read expected output file
	expectedOutputFile := filepath.Join(getDirPathFromDirName(dirName), expectedOutputFileName)
	expectedOutput, err := os.ReadFile(expectedOutputFile)
	require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
	actualOutputFile := filepath.Join(getDirPathFromDirName(dirName), actualOutputFileName)
	if string(expectedOutput) != actualOutput {
		// generate actual output file		
		fp, err := os.Create(actualOutputFile)
		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
		_, err = fp.WriteString(actualOutput)
		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
	}
	require.Equal(t, string(expectedOutput), actualOutput, "output mismatch for %s, actual output file %q vs expected output file: %q", getDebugMsgWithTestNameAndFormat(testName, format),
		actualOutputFile, expectedOutputFile)
}
```
Be consistent with `require` statements to fail the test rather than `t.Fatalf`.

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,44 +1,104 @@
...
+func checkErrorContainment(t *testing.T, testName, expectedErrorMsg, actualErrMsg string, isErr bool) {
+	errType := "error"
+	if !isErr {
+		errType = "warning"
+	}
+	if !strings.Contains(actualErrMsg, expectedErrorMsg) {
+		t.Fatalf("%s message mismatch for test %q, actual: %q, expected contains: %q", errType, testName, actualErrMsg, expectedErrorMsg)
+	}
+	require.Contains(t, actualErrMsg, expectedErrorMsg, "test: %q", testName)
+}
```
adisos, line 86: same here - no need for both `t.Fatalf` and `require.Contains` statements -- you can move the detailed str from the call of `Fatalf` to be the detailed str by the call to `require.Contains`

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,44 +1,104 @@
...
+
+func checkExtractedLineContainment(t *testing.T, testName, extractedLine, actualOutput string) {
+	if strings.Contains(actualOutput, extractedLine) {
+		t.Fatalf("test: %q, output should not contain %q", testName, extractedLine)
+	}
+	require.NotContains(t, actualOutput, extractedLine, "test: %q", testName)
```
adisos: and same here

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: eliminated all t.fatalf (my mistake : thought the msg in require would be printed always - so kept as test name and format only)

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1,44 +1,96 @@
...
+	expectedOutputFile := filepath.Join(getDirPathFromDirName(dirName), expectedOutputFileName)
+	expectedOutput, err := os.ReadFile(expectedOutputFile)
+	require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
+	actualOutputFile := filepath.Join(getDirPathFromDirName(dirName), actualOutputFileName)
+	if string(expectedOutput) != actualOutput {
+		// generate actual output file
+		fp, err := os.Create(actualOutputFile)
+		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
+		_, err = fp.WriteString(actualOutput)
+		require.Nil(t, err, getDebugMsgWithTestNameAndFormat(testName, format))
```
adisos, line 61: There is already a function `writeRes` in `eval_test` for writing to a file. Consider moving it to a common pkg and use the same function here as well, rather than having multiple variations for this code.

shireenf-ibm: done some changes.
more updates in `eval_test.go` will be done on the eval test improvements branch

pkg/netpol/common/netpol_testing_common.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// GenerateActualOutputFile generates actual output file
+func GenerateActualOutputFile(t *testing.T, testName, output, filePath string) {
+	fp, err := os.Create(filePath)
+	require.Nil(t, err, "test %q: error creating file %q", testName, filePath)
+	_, err = fp.WriteString(output)
+	require.Nil(t, err, "test %q: error writing output to file %q", testName, filePath)
+}
```
adisos, line 16: I think we can have a more general `func` for just writing to file.
No need for the `require` statements here, the test should not fail due to issue in writing to file.
It can return an `error` value, and in the tests you can print to log that there was issue in writing to file, if `err` isn't `nil`.
This function does not have to be specific for tests.

pkg/netpol/connlist/connlist_test.go
adisos: Note that I've updated the issue description - with list of packages to handle and priorities.
So I suggest waiting with `eval` pkg tests.

shireenf-ibm:
[this commit](https://github.com/np-guard/netpol-analyzer/pull/250/commits/e4d252c32960941d2f40e9c7486953e7b6d233f7)
fixes the lint issue on master branch

adisos (approved):

PR #258, opened 2023-10-31
cmd package test improvements

shireenf-ibm:
issue #110 

- tests improvements for cmd pkg :
- in this pkg could not run tests in parallel - seems that `cmd.Execute` func can not be run in parallel (when try to run in parallel, tests fail  , expected err/ output of a test returned as the actual of another)


shireenf-ibm:
hi @adisos , 
I think this PR is ready for review (although I am still learning the new refactoring changes)

-- on [commit ](https://github.com/np-guard/netpol-analyzer/pull/258/commits/486b3bf65f112cc817f55dbb8c7cde667f57ec48) , I had to delete a test with `json output `, because its `expected output file` caused two other tests in `connlist_test.go` to fail (it tries to read the json output file as an input file for those tests and fails with an error message containing `json: cannot unmarshal array into Go value of type unstructured.detector"`

cmd/netpolicy/cmd/command_test.go
```diff
@@ -2,159 +2,162 @@ package cmd
...
+	require.Equal(t, string(expectedOutput), actualOutput,
+		"output mismatch for test %q, actual output file %q vs expected output file: %q",
+		testName, actualOutputFile, expectedOutputFile)
+	if outputFile != "" {
+		_, err := os.Stat(outputFile)
+		require.Nil(t, err, "test: %q", testName)
+		fileContent, err := os.ReadFile(outputFile)
+		require.Nil(t, err, "test: %q", testName)
+		require.Equal(t, string(expectedOutput), string(fileContent),
+			"output mismatch for test %q, actual output file %q vs expected output file: %q", testName, outputFile, expectedOutputFile)
```
adisos, line 92: is this function similar to the one from `pkg/internal/testutils`? can the code duplication be avoided?

adisos (commented):
A few general comments:
-  Change Makefile to support the `update` flag for this pkg as well 
- Move json expected output file to some other place so that it does not interfere with the test you had to comment-out?

cmd/netpolicy/cmd/command_test.go
shireenf-ibm: yes both are very similar (only the part of `if outputFile != ""` is relevant for here only)
I think we can avoid code duplications by moving `internal/testutils` to be sibling of both `cmd` and `pkg`. Is this possible?
(this also true for `getTestsDir`)

adisos: Why not moving `pkg/internal/testutils` to be  `pkg/testutils`, and export this function, so that it can also be used under `cmd` tests?

shireenf-ibm: sure we can. (I thought `testutils ` was intended to be internal only)

shireenf-ibm:
> A few general comments:
> 
>     * Change Makefile to support the `update` flag for this pkg as well
done [here](https://github.com/np-guard/netpol-analyzer/pull/258/commits/8757a786a1bc65943c6215528121acc81356e01b)


>     * Move json expected output file to some other place so that it does not interfere with the test you had to comment-out?

moving `json` expected output means changing the standard `expected output` file which its path is computed in the same way for all other tests (<the common file prefix>.format under the test dir)
so this is a "special case" handled it in this [commit ](https://github.com/np-guard/netpol-analyzer/pull/258/commits/f20ee2ba1ebbf048df15b296515d205067cba6bf) by adding a flag for a "different" expected output path

cmd/netpolicy/cmd/command_test.go
shireenf-ibm: done

pkg/cli/command_test.go
```diff
@@ -244,14 +246,19 @@ func TestListCommandOutput(t *testing.T) {
...
+			specialOutFileFlag = true
+		}
 		t.Run(testName, func(t *testing.T) {
 			args := []string{"list", "--dirpath", filepath.Join(testutils.GetTestsDirWithDepth(currentDirDepth), tt.dirName)}
 			args = append(args, addCmdOptionalArgs(tt.format, tt.outputFile, tt.focusWorkload)...)
 			actualOut, err := buildAndExecuteCommand(args)
 			require.Nil(t, err, "test: %q", testName)
 			testutils.CheckActualVsExpectedOutputMatch(t, tt.dirName, expectedOutputFileName, actualOut, testInfo(testName), tt.outputFile,
-				currentDirDepth)
+				currentDirDepth, specialOutFileFlag)
```
adisos, line 269: I think we should avoid having "special flags", as this will be more confusing and difficult to maintain in the long term. Can you instead change the location for all expected output files, so that we have a uniform handling, and so that those output files would not interfere with the analysis results?

pkg/cli/command_test.go
```diff
@@ -2,580 +2,390 @@ package cli
...
-			isErr:          true, // fatal error because err is issued by the builder and uses stopOnErr
+			dir1:       "onlineboutique_workloads",
+			dir2:       "onlineboutique_workloads_changed_workloads",
+			format:     outpkg.TextFormat,
+			outputFile: outFileName,
 		},
+	}
+	for _, tt := range cases {
+		tt := tt
+		testName := testutils.DiffTestName(tt.dir1, tt.dir2) + "_format_" + determineFileSuffix(tt.format)
```
adisos, line 303: the format handling in the test name, should be part of the function `DiffTestName`

pkg/cli/command_test.go
```diff
@@ -2,580 +2,390 @@ package cli
...
+		t.Run(testName, func(t *testing.T) {
+			args := []string{"diff", "--dir1", filepath.Join(testutils.GetTestsDirWithDepth(currentDirDepth), tt.dir1), "--dir2",
+				filepath.Join(testutils.GetTestsDirWithDepth(currentDirDepth), tt.dir2)}
+			args = append(args, addCmdOptionalArgs(tt.format, tt.outputFile, "")...)
+			actualOut, err := buildAndExecuteCommand(args)
+			require.Nil(t, err, "test: %q", testName)
+			expectedOutputFileName := strings.Replace(testName, testutils.FormatStr, testutils.DotSign, 1)
```
adisos, line 310: add the generation of expected output file name to the function `DiffTestName`, similar to the `connlist` case

pkg/cli/command_test.go
```diff
@@ -2,580 +2,390 @@ package cli
...
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	outpkg "github.com/np-guard/netpol-analyzer/pkg/internal/output"
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/cli/command_test.go
```diff
@@ -2,580 +2,390 @@ package cli
...
+func testInfo(testName string) string {
+	return fmt.Sprintf("test: %q", testName)
 }
 
-const outFileName = "test_out.txt"
+// removes the output file generated for commands which run with `-f` flag
+func removeOutFile(t *testing.T, outputFile, testInfo string) {
+	if outputFile != "" {
+		err := os.Remove(outputFile)
+		require.Nil(t, err, testInfo)
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/cli/command_test.go
shireenf-ibm: done

pkg/internal/testutils/testutils.go
```diff
@@ -41,22 +46,38 @@ func GetTestsDirWithDepth(depth int) string {
 	return filepath.Join(res, testsDirName)
 }
 
-// GetDebugMsgWithTestNameAndFormat: testing helping func - writes debug message for good path tests
-func GetDebugMsgWithTestNameAndFormat(testName, format string) string {
-	return fmt.Sprintf("test: %q, output format: %q", testName, format)
+// ConnlistTestNameByTestType returns connlist test name and test's expected output file from some tests args
+func ConnlistTestNameByTestType(dirName, focusWorkload, format string) (testName, expectedOutputFileName string) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/internal/testutils/testutils.go
```diff
@@ -41,22 +46,38 @@ func GetTestsDirWithDepth(depth int) string {
...
+		return dirName + FormatStr + format, dirName + underscore + connlistExpectedOutputFilePartialName + format
+
+	case focusWorkload != "":
+		focusWorkloadStr := strings.Replace(focusWorkload, "/", underscore, 1)
+		namePrefix := dirName + focusWlAnnotation + focusWorkloadStr
+		return namePrefix + FormatStr + format,
+			namePrefix + underscore + connlistExpectedOutputFilePartialName + format
+	}
+	return "", ""
+}
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/internal/testutils/testutils.go
```diff
@@ -41,22 +46,38 @@ func GetTestsDirWithDepth(depth int) string {
...
+	expectedOutputFile := filepath.Join(GetTestsDirWithDepth(currDirDepth), outputFilesDir, testingPkg, expectedOutputFileName)
 	// if the --update flag is on (then generate/ override the expected output file with the actualOutput)
 	if *update {
 		err := output.WriteToFile(actualOutput, expectedOutputFile)
 		require.Nil(t, err, testInfo)
 		// if format is dot - generate/ override also png graph file using graphviz program
-		if strings.HasSuffix(expectedOutputFile, dotSign+output.DOTFormat) {
+		if strings.HasSuffix(expectedOutputFile, DotSign+output.DOTFormat) {
 			err = generateGraphFilesIfPossible(expectedOutputFile)
 			require.Nil(t, err, testInfo)
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/internal/testutils/testutils.go
```diff
@@ -66,7 +87,7 @@ func CheckActualVsExpectedOutputMatch(t *testing.T, testName, dirName, expectedO
 	expectedOutput, err := os.ReadFile(expectedOutputFile)
 	require.Nil(t, err, testInfo)
```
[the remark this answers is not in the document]
shireenf-ibm: done

adisos (changes requested):
few more comments

pkg/internal/testutils/testutils.go
```diff
@@ -41,42 +46,83 @@ func GetTestsDirWithDepth(depth int) string {
...
-	actualOutputFileName := "actual_" + expectedOutputFileName
-	actualOutputFile := filepath.Join(GetTestsDir(), dirName, actualOutputFileName)
+	if err != nil {
+		WarnOnErrorReadingFile(err, expectedOutputFile)
+	}
+	actualOutputFile := outFile
+	if outFile == "" {
+		actualOutputFileName := "actual_" + expectedOutputFileName
+		actualOutputFile = filepath.Join(GetTestsDirWithDepth(currDirDepth), outputFilesDir, testingPkg, actualOutputFileName)
+	}
```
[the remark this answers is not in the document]
shireenf-ibm: true; fixed

pkg/internal/testutils/testutils.go
```diff
@@ -41,42 +46,83 @@ func GetTestsDirWithDepth(depth int) string {
...
+func DiffTestNameByRefs(ref1, ref2 string) string {
+	return "diff_between_" + ref2 + "_and_" + ref1
+}
+
+// DiffTestNameByTestArgs returns diff test name and test's expected output file from some tests args
+func DiffTestNameByTestArgs(ref1, ref2, format string) (testName, expectedOutputFileName string) {
+	namePrefix := DiffTestNameByRefs(ref1, ref2)
+	testName = namePrefix + formatStr + format
+	expectedOutputFileName = namePrefix + dotSign + format
+	return testName, expectedOutputFileName
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/internal/testutils/testutils.go
```diff
@@ -41,42 +46,76 @@ func GetTestsDirWithDepth(depth int) string {
...
+// ConnlistTestNameByTestArgs returns connlist test name and test's expected output file from some tests args
+func ConnlistTestNameByTestArgs(dirName, focusWorkload, format string) (testName, expectedOutputFileName string) {
+	var namePrefix string
+	switch {
+	case focusWorkload == "":
+		namePrefix = dirName
+	case focusWorkload != "":
+		namePrefix = dirName + focusWlAnnotation + strings.Replace(focusWorkload, "/", underscore, 1)
+	}
+	testName = namePrefix + formatStr + format
```
adisos, line 58: can shorten this further:
```
namePrefix := dirName
if focusWorkload != "" {
  namePrefix += focusWlAnnotation + strings.Replace(focusWorkload, "/", underscore, 1)
}
```

pkg/internal/testutils/testutils.go
```diff
@@ -41,42 +46,73 @@ func GetTestsDirWithDepth(depth int) string {
...
+}
+
+// DiffTestNameByTestArgs returns diff test name and test's expected output file from some tests args
+func DiffTestNameByTestArgs(ref1, ref2, format string) (testName, expectedOutputFileName string) {
+	namePrefix := "diff_between_" + ref2 + "_and_" + ref1
+	testName = namePrefix
+	if format != "" {
+		testName += formatStr + format
+	}
+	expectedOutputFileName = namePrefix + dotSign + format
```
adisos, line 67: if `format` can be empty, how can a correct `expectedOutputFileName` be returned? 
cannot we assume `format` is not empty, and be consistent with `connlist` test names ?

adisos: The question is why is it used in `diff_test` with empty format? why there is no consistent use in `DiffTestNameByTestArgs` in all diff_tests? Can it be changed to be consistent?

pkg/netpol/diff/diff_test.go
```diff
@@ -29,7 +28,7 @@ func TestDiff(t *testing.T) {
 	t.Parallel()
 	for _, tt := range goodPathTests {
 		tt := tt
-		testName := getTestName(tt.firstDirName, tt.secondDirName)
+		testName, _ := testutils.DiffTestNameByTestArgs(tt.firstDirName, tt.secondDirName, "")
 		t.Run(testName, func(t *testing.T) {
```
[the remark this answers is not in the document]
adisos: Maybe move `t.Run()` to the inner level of the loops and rely on the value from `pTest` ?

adisos (changes requested):
Looks good!
One more comment: can you move the `output_files` dir to be not under `tests` dir (it is currently confusing as it is mixed with a long list of actual test dirs). Let it be a sibling of the `tests` dir, and rename to `test_outputs`.

adisos (approved):

Issue #264, opened 2023-11-08
move DiffAnalyzer to a separate file


adisos:
`DiffAnalyzer` deserves a file of its own - not a show stopper for the release, can wait for a follow-up PR

_Originally posted by @zivnevo in https://github.com/np-guard/netpol-analyzer/pull/261#discussion_r1386440467_

PR #270, opened 2023-11-12
moving diffAnalyzer to a new file

shireenf-ibm:
#264


pkg/netpol/diff/diff_analyzer.go
adisos, line 1: I would keep only lines 19-86 in the separate file for `diff_analyzer.go`, and the rest I think should remain in the original `diff.go` file.

shireenf-ibm: done

shireenf-ibm: shouldn't add also other exported funcs , like `NewDiffAnalyzer, and ConnectivityDiffToString`?

adisos: I meant lines 19-86 from the file diff_analyzer.go of your previous commit...
I think it should include the type `DiffAnalyzer` itself, and all the DiffAnalyzer options , including also `NewDiffAnalyzer`.
The rest (exported methods and non-exported methods and functions) in the original `diff.go` file.

shireenf-ibm: done

adisos (approved):

PR #271, opened 2023-11-12
Issue 265 moving pkgs

shireenf-ibm:
#265

in the meanwhile, kept `mnaifests_test.go` under `netpol` since it uses `netpol/internal/testutils` (till the decision if to move testutils, or maybe should move it to `pkg/internal/testutils`? (still will not be helpful for cmd pkg))


adisos:
> #265
> 
> in the meanwhile, kept `mnaifests_test.go` under `netpol` since it uses `netpol/internal/testutils` (till the decision if to move testutils, or maybe should move it to `pkg/internal/testutils`? (still will not be helpful for cmd pkg))

I would currently change `manifests_test.go` so that it does not depend on `testutils`, and keep it where it was.

shireenf-ibm:
> > #265
> > in the meanwhile, kept `mnaifests_test.go` under `netpol` since it uses `netpol/internal/testutils` (till the decision if to move testutils, or maybe should move it to `pkg/internal/testutils`? (still will not be helpful for cmd pkg))
> 
> I would currently change `manifests_test.go` so that it does not depend on `testutils`, and keep it where it was.

done

adisos (approved):

PR #273, opened 2023-11-14
rearrange packages

shireenf-ibm:
- `output formats`, `ingress pod consts` and `writeToFile` are/will be used by both `pkg/netpol` , `pkg/cli` and `pkg/internal/testutils` 
so moved them from old `pkg/netpol/common` to `pkg/internal/utils`


pkg/internal/utils/utils.go
```diff
@@ -29,10 +12,12 @@ const (
...
-// diff format common const
+// Ingress Controller const - the name and namespace of an ingress-controller pod
 const (
-	DotHeader  = "digraph {"
-	DotClosing = "}"
+	//  The actual ingress controller pod is usually unknown and not available in the input resources for the analysis.
+	// IngressPodName and IngressPodNamespace are used to represent that pod with those placeholder values for name and namespace.
+	IngressPodName      = "ingress-controller"
+	IngressPodNamespace = "ingress-controller-ns"
 )
```
adisos, line 21: why cannot this (`IngressPodName`, `IngressPodNamespace`) remain in `pkg/netpol/common` ? It seems to be used only by `connlist` and `diff`.

pkg/internal/utils/utils.go
```diff
@@ -1,24 +1,7 @@
...
-const (
-	//  The actual ingress controller pod is usually unknown and not available in the input resources for the analysis.
-	// IngressPodName and IngressPodNamespace are used to represent that pod with those placeholder values for name and namespace.
-	IngressPodName      = "ingress-controller"
-	IngressPodNamespace = "ingress-controller-ns"
-)
-
 // formats supported for output of various commands
 const (
 	DefaultFormat = "txt"
```
adisos, line 7: I think it is a bit confusing/weird that output formats are under `utils` pkg

adisos: Maybe split to `pkg/internal/output` and `pkg/internal/testutils` ?

pkg/internal/utils/utils.go
shireenf-ibm: in the new command_test version (#258), we have test with input `focusWorkload: "ingress-controller"`, so I am thinking assign it with the `IngressPodName const`

adisos: It is more related to packages under `pkg/netpol`, so I would keep it there. The api-based tests can use this const, and the cli test can just use this as hard-coded string.

pkg/internal/utils/utils.go
shireenf-ibm: done

adisos (approved):

Issue #241, opened 2023-09-21
extend dot output for tests automation


adisos:
Add automation that checks if graphviz is installed , and allows generating the `png` file from the `dot` expected output files.

PR #274, opened 2023-11-16
Issue 241 generate also png for dot output

shireenf-ibm:
issue #241


pkg/internal/testutils/testutils.go
```diff
@@ -80,3 +86,24 @@ func CheckErrorContainment(t *testing.T, testInfo, expectedErrorMsg, actualErrMs
...
+	executableNameForGraphviz = output.DOTFormat
+	// graph suffix
+	graphSuffix = "png"
+	dotSign     = "."
+)
+
+// checks if "graphviz" executable exists, if yes runs a cmd to generates a png graph file from the dot output
+func generateGraphFileIfPossible(dotFilePath string) error {
+	// check if graphviz is installed to continue
+	if _, err := exec.LookPath(executableNameForGraphviz); err == nil {
```
adisos, line 101: add a warning msg if could not find `dot` executable to re-generate `png` files

pkg/internal/testutils/testutils.go
```diff
@@ -80,3 +86,24 @@ func CheckErrorContainment(t *testing.T, testInfo, expectedErrorMsg, actualErrMs
 	require.Contains(t, actualErrMsg, expectedErrorMsg, "err/warn message mismatch for test: %q, actual: %q, expected contains: %q",
 		testInfo, actualErrMsg, expectedErrorMsg)
 }
+
+const (
+	// the executable we need from graphviz is "dot"
+	executableNameForGraphviz = output.DOTFormat
+	// graph suffix
+	graphSuffix = "png"
```
adisos, line 94: consider also adding `svg` files

adisos (commented):
Hi @shireenf-ibm ,
I tried this locally by running `make test-update`, but the `png` files were not regenerated..
when running `which dot` from my shell, I do get its path, and am able to run `dot` directly from the shell..

shireenf-ibm:
> Hi @shireenf-ibm , I tried this locally by running `make test-update`, but the `png` files were not regenerated.. when running `which dot` from my shell, I do get its path, and am able to run `dot` directly from the shell..

I am not sure that fully understood the question.
for me the first time I ran `make test-update` I really saw that the files were regenerated (in the commit we see that the files were deleted and new ones replaced them),
after that only if I delete a file it will be generated by running the command; 
If it exists, I will not see a change as it is the same content (same happens for all formats))

adisos:
> > Hi @shireenf-ibm , I tried this locally by running `make test-update`, but the `png` files were not regenerated.. when running `which dot` from my shell, I do get its path, and am able to run `dot` directly from the shell..
> 
> I am not sure that fully understood the question. for me the first time I ran `make test-update` I really saw that the files were regenerated (in the commit we see that the files were deleted and new ones replaced them), after that only if I delete a file it will be generated by running the command; If it exists, I will not see a change as it is the same content (same happens for all formats))

I thought that the `png` files could not be compared by git, and should have been marked as changed when re-generated, even though had the same content. Now I see they're indeed also compared and identified as identical, so it seems to work ok on my env :)  thanks!

pkg/internal/testutils/testutils.go
```diff
@@ -90,20 +91,23 @@ func CheckErrorContainment(t *testing.T, testInfo, expectedErrorMsg, actualErrMs
...
-			return err
+		for _, graphSuffix := range graphsSuffixes {
+			graphFilePath := dotFilePath + dotSign + graphSuffix
+			cmd := exec.Command("dot", dotFilePath, "-T"+graphSuffix, "-o", graphFilePath) //nolint:gosec // nosec
+			if err := cmd.Run(); err != nil {
+				return err
+			}
 		}
 	}
+	logger.NewDefaultLogger().Warnf("dot executable of graphviz was not found. Output Graphs will not be generated")
```
adisos, line 111: missing an `else` before this warning?

shireenf-ibm: true , made small rearrange to avoid inner identations

adisos (approved):

Issue #279, opened 2023-11-23
Improve connectivity visualization graphs


adisos:
- [ ] Add a frame of namespace for all the workloads at the same namespace. Workload name can then be shortened and not include the namespace.

PR #280, opened 2023-11-27
Dot - group peers in namespace frames, connlist

shireenf-ibm:
issue : #279 

hi @adisos,

**in this PR:**
- improving connlist dot output graphs by adding ns frames.

**some questions :**

- should add peers' `kind` label to its name?
- should keep the `ingress_controller_ns`  label in its frame? what about `ipblocks` frames (empty ns)?
- to add ns frames to `diff` dot output files too?


adisos:
> issue : #279
> 
> hi @adisos,
> 
> **in this PR:**
> 
>     * improving connlist dot output graphs by adding ns frames.
> 
> 
> **some questions :**
> 
>     * should add peers' `kind` label to its name?
> 
>     * should keep the `ingress_controller_ns`  label in its frame? what about `ipblocks` frames (empty ns)?
> 
>     * to add ns frames to `diff` dot output files too?

Looks good!
- I would keep the `kind` in the peer's name
- I would remove frames from ipblocks and from `ingress-controller`
- let's wait with `diff` output for a separate branch/PR

pkg/netpol/connlist/conns_formatter_dot.go
```diff
@@ -0,0 +1,125 @@
...
+func addNsGroups(nsPeersMap map[string][]string) []string {
+	res := []string{}
+	// sort namespaces (map's keys) to ensure same output always
+	nsKeys := sortMapKeys(nsPeersMap)
+	// write ns groups
+	for _, ns := range nsKeys {
+		peersLines := nsPeersMap[ns]
+		sort.Strings(peersLines)
+		// create ns  subgraph cluster
+		nsLabel := strings.ReplaceAll(ns, "-", "_")
```
adisos, line 108: why do we need this `replace`? does it change an original ns name?

adisos (commented):
in a separate PR (open another issue) -- update the documentation (readme) of the `connlist` output, add an example of the generated connectivity graph, and specify the frames are representing the namespaces in the analyzed cluster.

pkg/netpol/connlist/conns_formatter_dot.go
shireenf-ibm: no, it does not change the ns name (and the original namespace name is displayed in the frame `nsLines = append(nsLines, "\t\tlabel=\""+ns+"\"", "\t}")` which is on line 111),

this `nsLabel with replace`  is only used for writing the `subgraph header` which exists only in the dot file and will not appear in the graph itself.
replaced since the `dot` format does not accept `-` in its sub-graphs names (headers) . (we'll see a format error in the dot editor)

for example running :
`./bin/k8snetpolicy list --dirpath tests/minimal_test_in_ns/ -o dot` 
we have a namespace `hello-world`
in the dot format we'll see :
```
subgraph cluster_hello_world {      --> subgraph header with the replace to avoid dot format error
                "workload-a[Deployment]" [label="workload-a[Deployment]" color="blue" fontcolor="blue"]
                "workload-b[Deployment]" [label="workload-b[Deployment]" color="blue" fontcolor="blue"]
                label="hello-world"    --> here is the actual namespace name (also displayed in the frame)
}
```

adisos, line 108:
```suggestion
		nsLabel := strings.ReplaceAll(ns, "-", "_") // dot format does not accept "-" in its sub-graphs names (headers)
```

adisos (approved):

PR #285, opened 2023-12-07
Improve test dir paths handling 

shireenf-ibm:
issue #286

### DIFF
--- a/pkg/cli/command_test.go
+++ b/pkg/cli/command_test.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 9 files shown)


pkg/cli/command_test.go
```diff
@@ -136,9 +136,9 @@ func TestCommandsFailExecute(t *testing.T) {
 			args: []string{
 				"diff",
 				"--dir1",
-				filepath.Join(testutils.GetTestsDir(currentPkg), "onlineboutique_workloads"),
+				filepath.Join(testutils.TestsDirPath, "onlineboutique_workloads"),
```
adisos, line 139: Instead of duplicating the code `filepath.Join(testutils.TestsDirPath, "some test dir name")`, export a function from `testutils` that gets the dir name and returns the path for this test.

adisos (changes requested):
Looks good!
Added one comment

pkg/cli/command_test.go

adisos (approved):

PR #293, opened 2024-01-08
 initial exposure analysis 

shireenf-ibm:


shireenf-ibm:
hi @adisos,
this is an initial commit on exposure analysis.

1. still didn't append the results of exposure analysis into the connlist's result
but print them (for debug) to os.Stdout 
(and so some command tests fail as the command-line tests compare the output with os.Stdout ) 

2. for now exposure analysis results contain :
- pods which are not protected in the cluster
- pods which are not protected for one direction Ingress/Egress)
- pods which are captured by policies and exposed to potential namespaces (any namespace/ namespaces with particular labels)

3. I wrote all the new functions in new files (each file in the relevant package)

4. would like to discuss with you if changes in functionality are needed and the next steps (some were added in the issue)

 - some output examples :
 
Screenshot 2024-01-08 182639


Screenshot 2024-01-08 182711

 
Screenshot 2024-01-08 182424



Screenshot 2024-01-08 182535

adisos:
Hi @shireenf-ibm , few things to consider about the implementation:
- can the analysis be made while iterating pods and policies for connectivity analysis, instead of adding a separate analysis only after the connectivity analysis is already done?
- another approach to consider: can we save some code by adding fake pods, representing "arbitrary namespace" and certain namespaces appearing in policies that do not have actual resources?  if an actual pod in the connectivity analysis is connected to such "fake pod" , it means that it is exposed to its representing entity (any namespace / certain namespace) ? (thus we may get "exposure results" using the "connectivity analysis" functionality instead of adding a separate analysis). 
- let's also consider how this would be extended for pods exposure analysis, without implementing yet.

PR #295, opened 2024-01-13
wip - defining an interface for exposure-analysis results

shireenf-ibm:
this PR's goal is to define and implement the new API for `ConnlistFromResourceInfos` and `ConnlistFromDirPath` 

sub tasks :

- [x] define an interface for the new `exposure-analysis` results (new returned values from both funcs above)

this PR will resolve part of [this](https://github.com/np-guard/netpol-analyzer/pull/294#discussion_r1448389237)

### DIFF
--- a/pkg/netpol/connlist/connlist.go
+++ b/pkg/netpol/connlist/connlist.go
@@ -47,15 +47,15 @@ type ConnlistAnalyzer struct {
 // The new interface
 // ConnlistFromResourceInfos returns the allowed-connections list from input slice of resource.Info objects,
 // and the list of all workloads from the parsed resources
-func (ca *ConnlistAnalyzer) ConnlistFromResourceInfos(info []*resource.Info) ([]Peer2PeerConnection, []Peer, error) {
+func (ca *ConnlistAnalyzer) ConnlistFromResourceInfos(info []*resource.Info) ([]Peer2PeerConnection, []Peer, []ExposedPeer, error) {
 	// convert resource.Info objects to k8s resources, filter irrelevant resources
 	objs, fpErrs := parser.ResourceInfoListToK8sObjectsList(info, ca.logger, ca.muteErrsAndWarns)
 	ca.copyFpErrs(fpErrs)
 	if ca.stopProcessing() {
 		if err := ca.hasFatalError(); err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
-		return []Peer2PeerConnection{}, []Peer{}, nil
+		return []Peer2PeerConnection{}, []Peer{}, []ExposedPeer{}, nil
 	}
 	return ca.connslistFromParsedResources(objs)
 }
@@ -68,7 +68,7 @@ func (ca *ConnlistAnalyzer) copyFpErrs(fpErrs []parser.FileProcessingError) {
 
 // ConnlistFromDirPath returns the allowed connections list from dir path containing k8s resources
 // and list of all workloads from the parsed resources
-func (ca *ConnlistAnalyzer) ConnlistFromDirPath(dirPath string) ([]Peer2PeerConnection, []Peer, error) {
+func (ca *ConnlistAnalyzer) ConnlistFromDirPath(dirPath string) ([]Peer2PeerConnection, []Peer, []ExposedPeer, error) {
 	rList, errs := fsscanner.GetResourceInfosFromDirPath([]string{dirPath}, true, ca.stopOnError)
 	// instead of parsing the builder's string error to decide on error type (warning/error/fatal-err)
 	// return as fatal error if rList is empty or if stopOnError is on
@@ -79,7 +79,7 @@ func (ca *ConnlistAnalyzer) ConnlistFromDirPath(dirPath string) ([]Peer2PeerConn
 			err := utilerrors.NewAggregate(errs)
 			ca.logger.Errorf(err, netpolerrors.ErrGettingResInfoFromDir)
 			ca.errors = append(ca.errors, parser.FailedReadingFile(dirPath, err))
-			return nil, nil, err // return as fatal error if rList is empty or if stopOnError is on
+			return nil, nil, nil, err // return as fatal error if rList is empty or if stopOnError is on
 		}
 		// split err if it's an aggregated error to a list of separate errors
 		for _, err := range errs {
@@ -181,23 +181,24 @@ func (ca *ConnlistAnalyzer) hasFatalError() error {
 	return nil
 }
 
-func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8sObject) ([]Peer2PeerConnection, []Peer, error) {
+func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8sObject) ([]Peer2PeerConnection, []Peer,
+	[]ExposedPeer, error) {
 	// TODO: do we need logger in policyEngine?
 	pe, err := eval.NewPolicyEngineWithObjects(objectsList, ca.exposureAnalysis)
 	if err != nil {
 		ca.errors = append(ca.errors, newResourceEvaluationError(err))
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	ia, err := ingressanalyzer.NewIngressAnalyzerWithObjects(objectsList, pe, ca.logger, ca.muteErrsAndWarns)
 	if err != nil {
 		ca.errors = append(ca.errors, newResourceEvaluationError(err))
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 	return ca.getConnectionsList(pe, ia)
 }
 
 // ConnlistFromK8sCluster returns the allowed connections list from k8s cluster resources and a list of all peers names
-func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, []Peer, error) {
+func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, []Peer, []ExposedPeer, error) {
 	pe := eval.NewPolicyEngine(ca.exposureAnalysis)
 
 	// get all resources from k8s cluster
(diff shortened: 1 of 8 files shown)


shireenf-ibm:
hi @adisos ,
I had few ideas on how to write the interface of an `ExposedPeer` - (to add a `[]ExposedPeer` as the exposure-analysis results for the `ConnlistFromResourceInfos `and ConnlistFromDirPath functions.)
Chose the one in `pkg/netpol/connlist/exposed_pods.go`
would like to get your opinion before proceeding with computations.

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -0,0 +1,16 @@
...
+// ExposedPeer captures potential ingress and egress connections data for an exposed Peer
+type ExposedPeer interface {
+	ExposedPeer() Peer // UnprotectedPeer()
+	IngressExposure() []XgressExposureData
+	EgressExposure() []XgressExposureData
+}
+
+// XgressExposureData data of potential connectivity for an existing peer to/from a representative peer
+type XgressExposureData interface {
+	RepresentativePeer() Peer
```
adisos, line 14: `RepresentativePeer` is an internal implementation detail, and not something to expose..  I'd like to expose instead the actual relevant details: exposed namespace by labels / exposed workloads by labels / is exposed to entire cluster  (all namespaces) / is not protected by any netpol (per each direction) ? 

For the implementation of course, the `RepresentativePeer` will have to implement this api.

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -0,0 +1,16 @@
+package connlist
+
+import "github.com/np-guard/netpol-analyzer/pkg/netpol/internal/common"
+
+// ExposedPeer captures potential ingress and egress connections data for an exposed Peer
+type ExposedPeer interface {
+	ExposedPeer() Peer // UnprotectedPeer()
```
adisos, line 7:
```suggestion
	// ExposedPeer is a peer for which the analysis found some potential exposure info
	ExposedPeer() Peer 
```
We should watch the terms, as "unprotected" and "exposed" are not the same

pkg/netpol/connlist/connlist.go
```diff
@@ -47,15 +47,15 @@ type ConnlistAnalyzer struct {
...
 	// convert resource.Info objects to k8s resources, filter irrelevant resources
 	objs, fpErrs := parser.ResourceInfoListToK8sObjectsList(info, ca.logger, ca.muteErrsAndWarns)
 	ca.copyFpErrs(fpErrs)
 	if ca.stopProcessing() {
 		if err := ca.hasFatalError(); err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
-		return []Peer2PeerConnection{}, []Peer{}, nil
+		return []Peer2PeerConnection{}, []Peer{}, []ExposedPeer{}, nil
```
adisos, line 58: return `[]ExposedPeer{}` and not `nil` only if the flag for exposure analysis was true? (other places relevant)

pkg/netpol/connlist/connlist.go
```diff
@@ -47,15 +47,15 @@ type ConnlistAnalyzer struct {
 // The new interface
 // ConnlistFromResourceInfos returns the allowed-connections list from input slice of resource.Info objects,
 // and the list of all workloads from the parsed resources
```
adisos, line 49: update this comment, explain when `[]ExposedPeer` is returned as not just `nil`, and what it represents.

adisos (changes requested):

pkg/netpol/connlist/connlist.go
shireenf-ibm: sure, this is the plan, I just didn't give it attention in this commit , since my goal was only to consult about the interface 
(just wrote in a way that make build will not fail)

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -0,0 +1,27 @@
...
+	// IsProtectedByNetpols Is the exposed peer protected by any xgress netpol
+	IsProtectedByNetpols() bool
+	// IsExposedToEntireCluster is the peer exposed to all namespaces in the cluster for the relevant direction
+	IsExposedToEntireCluster() bool
+	// NamespaceLabels the potential namespaces which the peer is exposed to by their labels
+	NamespaceLabels() map[string]string
+	// PodLabels the potential pods which the peer might be exposed to by labels
+	PodLabels() map[string]string
+	// PotentialConnectivity the potential connectivity of the exposure
+	PotentialConnectivity() common.AllowedConnectivity
```
adisos, line 26: What combination of values is possible for these fields (each object is one instance in `IngressExposure()` or `EgressExposure()`) ?
How  is such an object going to be mapped to the visual connectivity output ? (what kind of edge is added and to which "new" node?)

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -0,0 +1,27 @@
...
+	ExposedPeer() Peer
+	// IngressExposure list of the potential Ingress connections to the ExposedPeer
+	IngressExposure() []XgressExposureData
+	// EgressExposure list of the potential Egress connections from the ExposedPeer
+	EgressExposure() []XgressExposureData
+}
+
+// XgressExposureData data of potential connectivity for an existing peer to/from a representative peer
+type XgressExposureData interface {
+	// IsProtectedByNetpols Is the exposed peer protected by any xgress netpol
```
adisos, line 17:
```suggestion
	// IsProtectedByNetpols indicates if the exposed peer is protected by any netpol on Ingress/Egress 
```
change documentation so that the first word is the field name, but also is the first word in the sentence (not a separate sentence that starts right after that) .

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: I don't think this is final, this may be changed during the computations, wanted to be aware of all data needed to be stored. 
( I am moving the `isProtected` flag out to the ExposedPod interface (splitted into 2 flags ingress+egress) - 

for case of non-protected peer (e.g. ingress not protected): ingress- exposure list will include one item with `exposedToEntireClass` and all-conns in the allowed connectivity >>
on the graph :  if the peer "free" (is not selected as "dst" for other peers in the graph (see question below)), will have an indication that it is exposed to any namespace (having a node for "any-namespace" and connected to it) 

but the classic case idea : 
let's assume we are having a pod `pod-a` which is selected by `ingress` netpols and having some potential conns (rules that has no match):
let's assume a rule with only-ns-selector, another with only-pod-selector and the third with both-ns-pod-selectors
it []ingress-exposure will be of len 3 :
ingress-exposure[0] will have non empty namespaceLabels with its allowed connectivity (protected, not exposed to entire cluster , empty podlabels)
ingress-exposure[1] will have non-empty podLabels
and ingress-exposure[2] will have non-empty nslabels and nonempty pod-labels 
in this case each item in the ingress-exposure list will have a node in the graph and edge to the `pod-a`

There are some questions to consider while computing: (maybe relevant to the "dot output" PR? or should be considered now?)
-> if the peer is protected (selected in ingress for example) and have an entire-cluster exposure (nsSelector : {} ), I think if there are other peers in the cluster with same connection , so automatically we will have ingress to it from other peers (should we also add a connection from "any namespace" node? )
-> same question if the peer is unprotected but have conns from other peers in the cluster?
-> if we have rule that exposes the peer to entire cluster and other rules with new-potential peers , i think we don't need to add node for the potential peers only the "any-namespace" node


- if we shouldn't add conns to/from entire cluster if there are already existing conns with other peers - I might also add a flags to indicate if there is connections between a peer and other peers in the cluster (with same allowed conns)

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: I am planning to implement `RepresentativePeer` in next PR
in this PR to focus on the implementations should be done to `connlist` and use the data I already have from the first (base of this) branch , do you agree?

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: - if a peer is exposed to entire cluster , it will have a representation of the connection/s with the entire cluster (i.e. "any-namespace" node) (even if it is connected to other peers in the cluster)
-  if we have rule that exposes the peer to entire cluster and other rules with new-potential peers - only the entire-cluster rule is considered (the most general connection)
- the new nodes types :
   - any-namespace
   - any namespace with labels
   - any pod with labels in the netpol's namespace
   - any pod with labels in any-namespace
   - any pod with labels in any namespace with labels

pkg/netpol/connlist/connlist.go
shireenf-ibm: will be resolved in next PR

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: done

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -6,21 +6,22 @@ import "github.com/np-guard/netpol-analyzer/pkg/netpol/internal/common"
...
-	// IngressExposure list of the potential Ingress connections to the ExposedPeer
+	// IngressExposure is a list of the potential Ingress connections to the ExposedPeer
 	IngressExposure() []XgressExposureData
-	// EgressExposure list of the potential Egress connections from the ExposedPeer
+	// EgressExposure is a list of the potential Egress connections from the ExposedPeer
 	EgressExposure() []XgressExposureData
 }
 
-// XgressExposureData data of potential connectivity for an existing peer to/from a representative peer
+// XgressExposureData contains data of potential connectivity for an existing peer to/from a representative peer
```
adisos, line 15: since this is the interface we should not use the internal term `representative peer` in the documentation here as well..

pkg/netpol/connlist/exposed_pods.go
adisos, line 1: be consistent with file name to the interface type it defines.. rename to "exposed_peer" ?

pkg/netpol/connlist/exposed_pods.go
```diff
@@ -0,0 +1,28 @@
...
+	// if a peer is not protected by xgress netpols, it will be exposed to entire cluster with all allowed connections
+	IsProtectedByNetpols() bool
+	// IsExposedToEntireCluster indicates if the peer is exposed to all namespaces in the cluster for the relevant direction
+	IsExposedToEntireCluster() bool
+	// NamespaceLabels are matchLabels of potential namespaces which the peer might be exposed to
+	NamespaceLabels() map[string]string
+	// PodLabels are matchLabels of potential pods which the peer might be exposed to
+	PodLabels() map[string]string
+	// PotentialConnectivity the potential connectivity of the exposure
+	PotentialConnectivity() common.AllowedConnectivity
```
adisos: is it possible to infer from the possible values of objects that implement this interface, the mapped node type in the planned output? 

what about the case of a specific namespace with specific pod labels (ns inferred from netpol)? how is it represented by such an object?
what about the case of specific pod lables with all namspaces? 

```
the new nodes types :

    any-namespace
    any namespace with labels
    any pod with labels in the netpol's namespace
    any pod with labels in any-namespace
    any pod with labels in any namespace with labels
```

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: changed

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: done

pkg/netpol/connlist/exposed_pods.go
shireenf-ibm: - for the specific namespace i think we may use the name selector of the namespace.
- for pods in any-namespace maybe if the namespaceLabels is empty (or with flag IsExposedToEntireCluster() true)

adisos: ok, should also update documentation to explain this mapping, so that one can interpret the objects correctly

pkg/netpol/connlist/exposed_peer.go
```diff
@@ -0,0 +1,30 @@
...
+	ExposedPeer() Peer
+	// IngressExposure is a list of the potential Ingress connections to the ExposedPeer
+	IngressExposure() []XgressExposureData
+	// EgressExposure is a list of the potential Egress connections from the ExposedPeer
+	EgressExposure() []XgressExposureData
+}
+
+// XgressExposureData contains the data of potential connectivity for an existing peer in the cluster
+// a peer might be exposed to the entire cluster (any-namespace), to any namespace with labels or
+// any pod with labels in any-namespace, or any pod with labels in a namespace with labels, or any pod with labels in a specific namespace
```
adisos, line 17:
```suggestion
// XgressExposureData contains the data of potential connectivity for an existing peer in the cluster
// a peer might be exposed to the entire cluster (any-namespace), to any namespace with labels or
// any pod with labels in any-namespace, or any pod with labels in a namespace with labels, or any pod with labels in a specific namespace
// TODO: add detailed documentation as to which combinations of values represent which kind of "abstract" node in the output
```

pkg/netpol/connlist/exposed_pods.go
adisos: > for the specific namespace i think we may use the name selector of the namespace.
I think this way we wouldn't know whether we can specify the namespace by name or by labels, but we can handle this at a later stage (ns by labels is originated from selector, ns by name is originated by netpol ns ... )

adisos (approved):

Issue #236, opened 2023-09-11
Exposure analysis


adisos:
Consider enhancement of the connectivity analysis: 
A "what-if" analysis, to report if an analyzed workload is exposed (ingress of egress) to workloads which are not included in the input manifests.

adisos:
More details:

Possible situations, which may not be reflected on the current connectivity report (since such connections may only be possible with peer workloads that do not exists on the input dir of the analyzed workload manifests):
- Pod is exposed to entire cluster (both ingress and egress) – not protected by netpols
- Pod is exposed to cluster only on ingress / egress (protected only on one direction by netpols) 
- Pod is exposed to potential specific namespaces / specific pods (which may exist or not on the cluster, but do not exist on the analyzed workload manifests) 

How to provide/add this information in a connectivity report / graph? 
(Currently, the concept of a “connection” between workloads in the report is defined only for workloads that exist on the input dir, and for which both source is allowed by netpols to egress to destination, and destination is allowed by netpols to ingress from source).

zivnevo:
The analysis provided by ACS only includes the first two bullets + indicating pods that are externally exposed.
![image](https://github.com/np-guard/netpol-analyzer/assets/79099626/1fee0bfb-144a-4824-bf77-a0e41967612b)

shireenf-ibm:
work plan:

- [ ] initial computing of exposure analysis - (basic exposure cases - entire class + rules with `namespaceSelector`)
- [ ] adding flag `--exposure-analysis` to turn on this feature only when needed 
- [ ] adding exposure analysis results to graphical output
- [ ] adding exposure analysis results to textual output
- [ ] expanding the exposure analysis to rules with `PodSelector` 
- [ ] adding more complicated tests

shireenf-ibm:
### Summary of the implementation flow: (highlights)

- exposure-analysis results are produced by running list command with `--exposure` flag
(`./bin/k8snetpolicy list --dirpath <testing-dir> --exposure`)

- the string output when running with exposure-analysis will include both 
	* connlist results, i.e. connections between peers from the manifests / connections between real peers and ip-blocks.
	* exposure analysis results, i.e. connections between real peers from the manifests and potential pods. (connections to any namespace in the cluster will be represented by `entire-cluster`, connections to external resources will be represented by ip-blocks).

[...]

PR #298, opened 2024-01-18
Revert api changes

shireenf-ibm:
issue #236 

sub task:

> *  - [ ] revert changes to connlist API funcs, add new func to get the []ExposedPeer
>              Also, do not change types for existing APIs

### DIFF
--- a/pkg/netpol/connlist/connlist.go
+++ b/pkg/netpol/connlist/connlist.go
@@ -191,10 +191,16 @@ func (ca *ConnlistAnalyzer) hasFatalError() error {
 
 func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8sObject) ([]Peer2PeerConnection, []Peer, error) {
 	// TODO: do we need logger in policyEngine?
-	pe, err := eval.NewPolicyEngineWithObjects(objectsList, ca.exposureAnalysis)
+	pe, err := eval.NewPolicyEngineWithObjects(objectsList)
 	if err != nil {
 		ca.errors = append(ca.errors, newResourceEvaluationError(err))
 		return nil, nil, err
+	}
+	if ca.exposureAnalysis {
+		err = pe.SetExposureAnalysisResources()
+		if err != nil {
+			return nil, nil, err
+		}
 	}
 	ia, err := ingressanalyzer.NewIngressAnalyzerWithObjects(objectsList, pe, ca.logger, ca.muteErrsAndWarns)
 	if err != nil {
@@ -205,8 +211,15 @@ func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8
 }
 
 // ConnlistFromK8sCluster returns the allowed connections list from k8s cluster resources, and list of all peers names
+// if exposure analysis option is on , performs exposure analysis computations
 func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, []Peer, error) {
-	pe := eval.NewPolicyEngine(ca.exposureAnalysis)
+	pe := eval.NewPolicyEngine()
+	if ca.exposureAnalysis {
+		err := pe.SetExposureAnalysisResources()
+		if err != nil {
+			return nil, nil, err
+		}
+	}
 
 	// get all resources from k8s cluster
 
--- a/pkg/netpol/connlist/connlist.go
+++ b/pkg/netpol/connlist/connlist.go
@@ -165,9 +165,9 @@ func (ca *ConnlistAnalyzer) Errors() []ConnlistError {
 	return ca.errors
 }
 
-// ExposureAnalysis returns a slice of ExposedPeer objects,  capturing for input workloads if/how they  may
+// ExposedPeers returns a slice of ExposedPeer objects,  capturing for input workloads if/how they  may
 // be exposed to potential cluster entities, based on the input network policies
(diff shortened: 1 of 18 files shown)


shireenf-ibm:
the only exposed function that I did not revert its changes :
`func (pe *PolicyEngine) AddPodByNameAndNamespace(name, ns string, nsLabels map[string]string) (Peer, error) `

it is exposed for `eval` pkg but used in internal function in `connlist` 
so I could not decide if it satisfies the requirement of a new major version!!

pkg/netpol/connlist/exposed_peer.go
```diff
@@ -34,5 +32,5 @@ type XgressExposureData interface {
 	// PodLabels are matchLabels of potential pods which the peer might be exposed to
 	PodLabels() map[string]string
 	// PotentialConnectivity the potential connectivity of the exposure
-	PotentialConnectivity() conn.AllowedSet
+	PotentialConnectivity() common.Connection
```
adisos, line 35: I'm not sure want the return value to be an interface type from internal package. 
Please add another sub-task to address this later.

pkg/netpol/connlist/connlist.go
```diff
@@ -176,6 +165,11 @@ func (ca *ConnlistAnalyzer) Errors() []ConnlistError {
 	return ca.errors
 }
 
+// ExposureAnalysis returns a slice of ExposedPeer with all exposed pods and their exposure data
```
adisos, line 168:
```suggestion
// ExposureAnalysis returns a slice of ExposedPeer objects,  capturing for input workloads if/how they  may
// be exposed to potential cluster entities, based on the input network policies
```

pkg/netpol/connlist/connlist.go
```diff
@@ -176,6 +165,11 @@ func (ca *ConnlistAnalyzer) Errors() []ConnlistError {
...
+func (ca *ConnlistAnalyzer) ExposureAnalysis() []ExposedPeer {
```
adisos, line 169: consider renaming function to `ExposedPeers()` instead

pkg/netpol/connlist/connlist.go
```diff
@@ -195,27 +189,37 @@ func (ca *ConnlistAnalyzer) hasFatalError() error {
...
 	return ca.getConnectionsList(pe, ia)
 }
 
-// ConnlistFromK8sCluster returns the allowed connections list from k8s cluster resources, a list of all peers names,
-// and list of exposed peers in the parsed resources and their potentially allowed connections
-// if exposure-analysis option is on, otherwise nil
-func (ca *ConnlistAnalyzer) ConnlistFromK8sCluster(clientset *kubernetes.Clientset) ([]Peer2PeerConnection, []Peer, []ExposedPeer, error) {
-	pe := eval.NewPolicyEngine(ca.exposureAnalysis)
+// ConnlistFromK8sCluster returns the allowed connections list from k8s cluster resources, and list of all peers names
+// if exposure analysis option is on , performs exposure analysis computations
```
adisos, line 214:
```suggestion
```

pkg/netpol/eval/exposure.go
```diff
@@ -26,8 +26,8 @@ import (
 
 // this file contains eval.PolicyEngine funcs which are related to exposure-analysis feature
 
-// addPotentiallyExposedPods adds pods which are potentially exposed to connect with (from/to) existing pods in the resources
-func (pe *PolicyEngine) addPotentiallyExposedPods() error {
+// addInferredFromPoliciesPods adds representative pods which are exposed to connect with (from/to) existing pods in the resources
+func (pe *PolicyEngine) addInferredFromPoliciesPods() error {
```
adisos, line 30: rename func to `addRepresentativePods`

pkg/netpol/eval/exposure.go
```diff
@@ -26,8 +26,8 @@ import (
 
 // this file contains eval.PolicyEngine funcs which are related to exposure-analysis feature
 
-// addPotentiallyExposedPods adds pods which are potentially exposed to connect with (from/to) existing pods in the resources
-func (pe *PolicyEngine) addPotentiallyExposedPods() error {
+// addInferredFromPoliciesPods adds representative pods which are exposed to connect with (from/to) existing pods in the resources
```
adisos, line 29:
```suggestion
// addInferredFromPoliciesPods adds representative pods by inferring from network policies selectors within rules
// the required entities to represent (namespaces or pods by certain labels selectors)
// for example, if a rule within policy has namespace selector "name: foo", then a representative pod in such a 
// namespace with those labels will be added, representing all potential pods in such a namespace
```

adisos (changes requested):
added a few comments.
you can keep the func's `AddPodByNameAndNamespace` new version as it is.

pkg/netpol/eval/exposure.go

pkg/netpol/connlist/connlist.go

pkg/netpol/connlist/exposed_peer.go
shireenf-ibm: a solution I could think of is to expose common.Connection as connlist.AllowedSet ( so we don't duplicate the interface implementation again)  - done on [commit](https://github.com/np-guard/netpol-analyzer/pull/298/commits/bd98185794a50023ff2d74db9c1307007dd681aa)

adisos: let's revert this commit and think about other options later.. should be done in a separate sub task

adisos (approved):

PR #296, opened 2024-01-16
connlist implementing exposure analysis

shireenf-ibm:
issue #236

task + sub-task :

```
- [on base branch ] initial computing of exposure analysis - (basic exposure cases - entire class + rules with `namespaceSelector`)
   *  - [ this branch] code implementations in connlist: compute (separate) exposure-analysis results in `getConnectionsBetweenPeers` into the new returned value
```

the focus in this PR is to store the exposure analysis conns separately from the p2pconns; i.e implementations in `connlist` pkg

**_main file changes in `connlist`:_**
- **pkg/netpol/connlist/connlist.go :** 
    *_relevant to review:_ 
      - the functions at the end of the file - splitting exposure conns and p2pconns , and storing the exposure analysis conns in a map , and the call from and `getConnectionsBetweenPeers`, and `getConnectionsList`
      * _will be changed in next PRs:_ 
       - all the returned values `[]ExposedPeer` (in the API functions, and etc.), `emptyExposedListOrNil()`  : will be reverted in next PR (next sub task), 
       - checking if a peer is representative by comparing its name : will be enhanced in a later PR (implementing RepresentativePeer`

- **pkg/netpol/connlist/exposed_peer.go**: (changes are relevant to review) 
- **pkg/netpol/connlist/exposure_analysis.go** : all the file is relevant to review, implementing the interfaces from `pkg/netpol/connlist/exposed_peer.go`  + functionality to compute `[]ExposedPeer` from the map computed in `connlist.go`
       
      

main changes under `eval` pkg which are relevant to review are related to the new `pod` flags : `IngressProtected and EgressProtected` (adding + handling them), + removed unused code

PR #304, opened 2024-01-26
Entire cluster exposure opt

shireenf-ibm:
issue #236 

sub-task: 

>  *  - [ ] save on the fly - if a peer is exposed to entire cluster and the largest connection-set the allows such connectivity, and remove fake pod for "any namespace"  [more info](https://github.com/np-guard/netpol-analyzer/pull/296#discussion_r1454893972) and [here](https://github.com/np-guard/netpol-analyzer/pull/296#discussion_r1455117402)


pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -182,12 +193,12 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
...
 				selector, err := np.parseNetpolLabelSelector(rulePeers[i].PodSelector)
 				if err != nil {
-					return false, err
+					return false, false, err
 				}
 				peerMatchesPodSelector = selector.Matches(labels.Set(peer.GetPeerPod().Labels))
 			}
 			if peerMatchesPodSelector {
-				return true, nil //  matching both pod selector and ns_selector here
+				return true, anyNsMatch, nil //  matching both pod selector and ns_selector here
```
adisos, line 201: if the function returns here on the first rule (which is not an "all ns" match), we can miss an "all ns" match if it is in the second rule?

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -144,32 +144,43 @@ func (np *NetworkPolicy) ruleConnsContain(rulePorts []netv1.NetworkPolicyPort, p
 }
 
 // ruleSelectsPeer returns true if the given peer is in the set of peers selected by rulePeers
+// also returns an indication if the rulePeers matches all namespaces in the cluster
 //
 //gocyclo:ignore
-func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, peer Peer) (bool, error) {
+func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, peer Peer) (peerSelected, anyNsMatch bool, err error) {
+	entireClusterCheckOnly := false
```
adisos, line 151: the var `entireClusterCheckOnly` and its logic should be better documented

adisos (commented):
a few initial comments and questions, did not look at all files yet.

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: seemed to me, that always the (on first call) first peer sent to this function (to be checked if selected) is the `IP-block peer` - and whether it is selected or not, I continued to check for "all ns" with the `entireClusterCheckOnly` flag.

(from this assumption I didn't add the same check with `entireClusterCheckOnly` flag here too.); shouldn't I assume the IP-block will be always sent first?

shireenf-ibm: but if first rule is `from IPBlock` , even though the peer is IP-type; it will miss the `all ns` check (so at least on that return satement `line 217` I need to fix )

shireenf-ibm: and will not assume first peer is always IP (and fix for this rule too `line 201`)

adisos: I think this function implementation might be having too many assumptions, causing this to be error prone.. I wouldn't rely on having a call with an IP peer, and definitely not on the order.

adisos: I'd consider a separate function to check policy rules only once for this specific goal, on the first time we start analyzing this policy (store the results so we do not repeat those computations).

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: not relevant anymore

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -42,8 +42,22 @@ type Pod struct {
...
-	EgressProtected  bool // indicates if the pod is selected by any egress netpol
+	// Following fields contain data on pod's exposure on the cluster (e.g. is protected by netpols?! is exposed to entire cluster?!
+	//	on what directions? on which connections? etc.)
+	IngressProtected               bool                  // indicates if the pod is selected by any ingress netpol
+	IngressExposedToEntireCluster  bool                  // indicates if ingress netpols rules expose the pod to entire cluster
+	IngressEntireClusterConnection *common.ConnectionSet // contains the maximum range of connection which exposes the
+	// pod to entire cluster on ingress
+	IngressToEntireClusterChecked bool // indicates if the ingress conn to entire cluster already checked, it is enough to check once;
+	// at the first time the ingress policies selecting the pod are looped
+
```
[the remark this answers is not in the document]
shireenf-ibm: fixed all

pkg/netpol/eval/internal/k8s/netpol.go
adisos: not sure I understand (2) - can you reference to an example?

shireenf-ibm: suupose we have following netpol:

> 
```
kind: NetworkPolicy
apiVersion: networking.k8s.io/v1
metadata:
  name: allow-hello-world-b-to-a-app
  namespace: hello-world
spec:
  podSelector:
    matchLabels:
      app: a-app
  policyTypes:
  - Ingress
  - Egress
  ingress:
  - from:
    - namespaceSelector: {}
    ports:
    - port: 8050
      protocol: TCP
  egress:
  - to:
    - namespaceSelector: {}
```

and only one peer in the resources `deployment with app: a-app`.

the only peers we'll get for looping in the `connlist.getConnectionsBetweenPeers` are the `a-app deployment and full IP-block`.
since the netpol restricts the conns only in the cluster; the allowed connections computed between the a-peer an ip-peer
(`in connlist line 504: allowedConnections, err := pe.AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer))` 
will be empty; and hence we'll not get to the call in `connlist 518 : p2pConnection, err := ca.checkIfP2PConnOrExposureConn(pe, allowedConnections, srcPeer, dstPeer, exposuresMap)`. 
so I had to update the `a-app peer` entire cluster connections; also when allowed conns between "existing" peers are empty. 
(so I had to add the check in `connlist lines 508-513 calling updatePeersGeneralExposureData(pe, srcPeer, dstPeer, ingressSet, egressSet, exposuresMap)`

adisos (commented):
few comments, did not review all files yet

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -252,3 +266,20 @@ func (pod *Pod) ConvertPodNamedPort(namedPort string) int32 {
...
 }
+
+// updatePodXgressExposureToEntireClusterData updates the pods' fields which are related to entire class exposure on ingress/egress
+func (pod *Pod) updatePodXgressExposureToEntireClusterData(ruleConns *common.ConnectionSet, isIngress bool) {
+	if isIngress {
+		if pod.IngressEntireClusterConnection == nil {
+			pod.IngressEntireClusterConnection = ruleConns
+		} else {
+			pod.IngressEntireClusterConnection.Union(ruleConns)
+		}
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -42,8 +42,18 @@ type Pod struct {
...
-	EgressProtected  bool // indicates if the pod is selected by any egress netpol
+	// IngressProtected indicates if the pod is selected by any network-policy on ingress direction, or not
+	IngressProtected bool
+	// IngressEntireClusterConnection contains the maximum range of connections which the pod is exposed with to all namespaces
+	// on ingress direction; or nil if pod is not exposed to all namespaces on ingress direction
+	IngressEntireClusterConnection *common.ConnectionSet
+
+	// EgressProtected indicates if the pod is selected by any network-policy on egress direction, or not
+	EgressProtected bool
+	// EgressEntireClusterConnection contains the maximum range of connections which the pod is exposed with to all namespaces
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -42,8 +42,18 @@ type Pod struct {
 	HostIP    string
 	Owner     Owner
 
-	IngressProtected bool // indicates if the pod is selected by any ingress netpol
-	EgressProtected  bool // indicates if the pod is selected by any egress netpol
+	// IngressProtected indicates if the pod is selected by any network-policy on ingress direction, or not
+	IngressProtected bool
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -407,3 +407,64 @@ func (np *NetworkPolicy) Selects(p *Pod, direction netv1.PolicyType) (bool, erro
...
+		}
+		src.GetPeerPod().updatePodXgressExposureToEntireClusterData(ruleConns, false)
+	}
+	return nil
+}
+
+// ruleSelectsEntireCluster checks if the given rules list is exposed to entire cluster;
+// i.e. if the rules list is empty or if there is a rule with empty namespaceSelector
+// since this func is definitely called after ruleSelectsPeer, there is no rules' correctness check here
+func (np *NetworkPolicy) ruleSelectsEntireCluster(rules []netv1.NetworkPolicyPeer) bool {
```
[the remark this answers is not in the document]
shireenf-ibm: it is possible to store per policy following data (for both ingress and egress )
```
allowsConnsToEntireWorld bool
connToEntireCluster *common.connectionSet
```
when a policy is added to the engine.

however I think this should be done in another PR (before optimizing generating representative pods)
because by scanning policy rules on while "building" the policy-engine for our required data, we should check the rules' correctness (legality)
this correctness check currently is done in `ruleSelectsPeer` (that's why called this after `ruleSelectsPeer` ) - but then correctness check will be eliminated from `ruleSelectsPeer`

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -42,8 +56,17 @@ type Pod struct {
 	HostIP    string
 	Owner     Owner
 
-	IngressProtected bool // indicates if the pod is selected by any ingress netpol
-	EgressProtected  bool // indicates if the pod is selected by any egress netpol
+	// IngressExposureData contains:
+	// - whether the pod is protected by any network-policy on ingress direction or not;
+	// - and the maximal connection-set for which the pod is exposed to all namespaces by network policies
+	// on ingress direction
+	IngressExposureData podExposureInfo
```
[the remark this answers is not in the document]
shireenf-ibm: I think this is a typo;
the fields of the struct `podExposureInfo`  are exported .
if we don't want to export them; I'll add  `getter` funcs to return their values (e.g `getPodIsProtected`) (as we want them in `eval\exposure.go)

pkg/netpol/eval/internal/k8s/netpol.go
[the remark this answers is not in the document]
shireenf-ibm: I see;
1. yes we can assume rules are valid in preprocessing;
another approach, if upserting policies for policy engine is the same also for `eval` command; we can always store the data for the policy (not only for exposure-analysis; and then will benefit from these data before scanning the policies) 
2. yes; if we'll store this data; while looping policies selecting a pod; we can update the pod's entireClusterConnection from this policy data (`ScanRulesForIngressFromEntireCluster` is deprecated in this case)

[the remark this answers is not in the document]
adisos: 1. it depends if when adding policies we activate the pre-processing only for exposure analysis..
2. right

pkg/netpol/eval/internal/k8s/pod.go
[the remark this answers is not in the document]
adisos: ok, it can be exported, but should be consistent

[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/connlist/connlist.go
```diff
@@ -501,6 +505,12 @@ func (ca *ConnlistAnalyzer) getConnectionsBetweenPeers(pe *eval.PolicyEngine, pe
 			if err != nil {
 				return nil, nil, err
 			}
+			if ca.exposureAnalysis {
+				err = updatePeersGeneralExposureData(pe, srcPeer, dstPeer, ingressSet, egressSet, exposuresMap)
```
adisos, line 509: can just call `addPeerGeneralExposure` on `src` peer for both ingress and egress (right after `src` is assigned), and then no need to have `ingressSet` and `egressSet`

pkg/netpol/eval/internal/k8s/netpol.go
adisos: ok, so both (1) and (2) are related to the same fix in `connlist`?

pkg/netpol/connlist/connlist.go
shireenf-ibm: we can not update the general exposure before calling `pe.AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer)` at least once; 
since those fields are updated on the fly during the computation of the allowed conns between the src and dst.(on the first time a src appears in this call: its egress fields are updated; and in the first call with the dst: its ingress fields updated)
and therefore we can not update both egress and ingress peer's fields at once (we need a pod to have both "attitudes" in order to be updated) ; thus we need both egress and ingress sets.

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: yes;
because of 2 ; i had to move the call (`updatePeersGeneralExposureData`) from `checkIfP2PConnOrExposureConn` to here.
and because of 1; i had to update also "unprotected" data in this call (not only entireClusterConns)

adisos (approved):

PR #306, opened 2024-02-05
pre proccessing policy for general conns

shireenf-ibm:
issue : #236
sub-task:
   * - [ ] optimize policy pre-processing : store data on policy's general conns (to whole world/ entire cluster)[more info](https://github.com/np-guard/netpol-analyzer/pull/304#discussion_r1471739581)

### DIFF
--- a/pkg/netpol/eval/internal/k8s/netpol.go
+++ b/pkg/netpol/eval/internal/k8s/netpol.go
@@ -133,7 +133,7 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
 		if rulePorts[i].Protocol != nil {
 			protocol = *rulePorts[i].Protocol
 		}
-		ports := common.PortSet{}
+		ports := common.MakePortSet(false)
 		if rulePorts[i].Port == nil {
 			ports = common.MakePortSet(true)
 		} else {
--- a/pkg/netpol/eval/internal/k8s/netpol.go
+++ b/pkg/netpol/eval/internal/k8s/netpol.go
@@ -46,9 +46,6 @@ type NetworkPolicy struct {
 	// - the maximal connection-set which the policy's rules allow to all destinations on egress direction
 	// - the maximal connection-set which the policy's rules allow to all namespaces in the cluster on egress direction
 	EgressGeneralConns PolicyGeneralRulesConns
-
-	namedPortsToPortNum map[v1.Protocol]map[string]int32 // internal map; map from protocol to map from namedPort which
-	// appeared in the policy rules to its matching port number
 }
 
 type PolicyGeneralRulesConns struct {
@@ -76,17 +73,6 @@ func getProtocolStr(p *v1.Protocol) string {
 
 func (np *NetworkPolicy) convertNamedPort(namedPort string, pod *Pod) int32 {
 	return pod.ConvertPodNamedPort(namedPort)
-}
-
-func (np *NetworkPolicy) saveNamedPortToPort(protocol v1.Protocol, namedPort string, portNum int32) {
-	if np.namedPortsToPortNum == nil {
-		np.namedPortsToPortNum = make(map[v1.Protocol]map[string]int32)
-	}
-	// a namedPort was converted, store in the map
-	if _, ok := np.namedPortsToPortNum[protocol]; !ok {
-		np.namedPortsToPortNum[protocol] = make(map[string]int32)
-	}
-	np.namedPortsToPortNum[protocol][namedPort] = portNum
 }
 
 // getPortsRange returns the start and end port numbers given input port, endPort and dest peer
@@ -119,7 +105,7 @@ func (np *NetworkPolicy) getPortsRange(port *intstr.IntOrString, endPort *int32,
 }
 
 func isEmptyPortRange(start, end int32) bool {
-	return start == noPort && end == noPort
+	return start == common.NoPort && end == common.NoPort
 }
 
 func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, dst Peer) (*common.ConnectionSet, error) {
@@ -142,11 +128,10 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
 				return res, err
 			}
 			if dst == nil && portName != "" {
-				// if dst is nil and rulePorts contains namedPort, we want to save the general connection with namedPort
+				// adding namedPort to connectionSet in case of :
+				// -  dst is nil - for general connections;
+				// - TODO: if dst is a fake pod (representative)
 				ports.AddPort(intstr.FromString(portName))
-			}
-			if portName != "" && startPort != -1 {
-				np.saveNamedPortToPort(protocol, portName, startPort)
 			}
 			if !isEmptyPortRange(startPort, endPort) {
 				ports.AddPortRange(int64(startPort), int64(endPort))
@@ -535,23 +520,4 @@ func (np *NetworkPolicy) doRulesExposeToAllDestOrEntireCluster(rules []netv1.Net
 		}
 	}
 	return false, false
-}
-
-// UpdatePodEntireClusterConnFromPolicyData updates the pod's entire cluster connections on the given direction
-func (np *NetworkPolicy) UpdatePodEntireClusterConnFromPolicyData(podPeer Peer, isIngress bool) {
-	if isIngress {
-		// if ingress: the pod is the dst; its named port should be converted to numbers;
-		// if the policy's entire cluster connection includes
-		// named ports; add the matching port num to the connection
-		// should replace namedport here or move to another place?
-		np.IngressGeneralConns.EntireClusterConns.ReplaceNamedPortWithMatchingPortNum(np.namedPortsToPortNum)
-		podPeer.GetPeerPod().updatePodXgressExposureToEntireClusterData(np.IngressGeneralConns.EntireClusterConns, isIngress)
-	} else {
-		// TBD : if egress, the pod is the src; if the policy's entire cluster connection includes named ports;
-		// should we keep the named port only?
-		// (for specific real pods; this port is converted when computing the allowed conns between both peers)
-		// should following line: adding already converted namedPorts be here:
-		np.EgressGeneralConns.EntireClusterConns.ReplaceNamedPortWithMatchingPortNum(np.namedPortsToPortNum)
-		podPeer.GetPeerPod().updatePodXgressExposureToEntireClusterData(np.EgressGeneralConns.EntireClusterConns, isIngress)
-	}
 }
--- a/pkg/netpol/eval/internal/k8s/netpol.go
+++ b/pkg/netpol/eval/internal/k8s/netpol.go
@@ -537,6 +537,7 @@ func (np *NetworkPolicy) doRulesExposeToAllDestOrEntireCluster(rules []netv1.Net
 	return false, false
 }
 
+// UpdatePodEntireClusterConnFromPolicyData updates the pod's entire cluster connections on the given direction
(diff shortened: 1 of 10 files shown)


shireenf-ibm:
Some notes:

_**1. main changes:**_

**- step1 :** while upserting a new network-policy, scan its ingress and egress rules and store : 

	- all destinations (whole world) connections
	- entire cluster connections
			
notes on step1: 

- can not do the pre-process only for exposure analysis; since policy-engine's exposure analysis flag is updated only after upserting the objects  (the `pe.exposureAnalysisFlag` is updated after `pe.NewPolicyEngineWithObjects` which calls `UpsertObject` funcs)
    

- currently, I stored both entire world and entire cluster conns (entire cluster contains also entire world) although I use only entire cluster;

    in the mean while kept the entire-world connection cause we may use it in next optimizations; if we don't will eliminate it.
    

- a rule may contain `namedPort`; in pre-process we can not convert the `namedPort` into `portNum` (we don't have all peers yet, neither know what peers are selected); so :
    1. on pre-process; I store the new connection with the `namedPort` (however; here I found a bug in `PortSet`(detailed below))
    2. while computing allowed connections between peers (scanning the policy rules for finding selected peers;), each time a namedPort was converted, saved it in the policy struct; (a policy object has a map from protocol to map from namedPorts to port numbers)
 

**- step2:** while scaning the policies selecting a pod for allowed connections computations; updated the pod's general exposure data from the policy's stored data.

 (when updating the pod's data, checked and replaced `namedPort` in the poliy's entire cluster conns using the map; (worked with local test with namedPort))
 

**_2. notes on `PortSet` :_**
- although `PortSet` has a 
```
NamedPorts         map[string]bool
ExcludedNamedPorts map[string]bool
```
they were never used before and never initialized;
so while running a test with a `namedPort` I got : `panic: assignment to entry in nil map` error;
I updated the `MakePortSet` with initialized empty maps; but still got the error; 
since almost always a new `PortSet` is initialized by `PortSet{}`;
so currently in the `PortSet funcs` that are used for storing the `namedPort`; I check and initialize the maps if are nil

shireenf-ibm:
** TODO:** (for next commit?)
if new data of the Policy is accepted :
- we can benefit from it for computing allowed conns;
I suggest that the netpol func `ruleSelectsPeer` checks only if there is a specified rule selecting the specific peer; (specified podSelector/ NamespaceSelector ...)
and  then for-example : 
the allowed conns between peers is the entireCluster.Union(this-rules-conn)
for IPblock the policy's entireWorldConn.Union(specific IPBlock rule conn)

adisos:
some questions/comments:

> can not do the pre-process only for exposure analysis; since policy-engine's exposure analysis flag is updated only after upserting the objects 

I think we can find solutions for this, such as adding more api funcs to generate policy engine as desired.

> while computing allowed connections between peers (scanning the policy rules for finding selected peers;), each time a namedPort was converted, saved it in the policy struct

why does it need to be stored in the policy?

what if the connection from actual pod to a "fake" (representative) pod depends on a named port? how will this be handled by exposure analysis? 

> I suggest that the netpol func ruleSelectsPeer checks only ...

not sure I understand the entire sentence, though I think `ruleSelectsPeer` could avoid iterating all rules in case we already know from pre-processing (when available, from exposure analysis activation) that at lease one of the rules is matching all namespaces or any pod/ip-block ..  (for this case we do need the "all destinations (whole world) connections").

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -29,14 +29,35 @@ import (
...
+	// IngressGeneralConns contains:
+	// - the maximal connection-set which the policy's rules allow to all destinations on ingress direction
+	// - the maximal connection-set which the policy's rules allow to all namespaces in the cluster on ingress direction
+	IngressGeneralConns PolicyGeneralRulesConns
+	// EgressGeneralConns contains:
+	// - the maximal connection-set which the policy's rules allow to all destinations on egress direction
+	// - the maximal connection-set which the policy's rules allow to all namespaces in the cluster on egress direction
+	EgressGeneralConns PolicyGeneralRulesConns
+
+	namedPortsToPortNum map[v1.Protocol]map[string]int32 // internal map; map from protocol to map from namedPort which
```
adisos, line 50: the matching of named port to its number is per pod, so why is it stored here?

shireenf-ibm: answered in the previous comment

shireenf-ibm: fixed the handling of named-ports

**adding some examples - locally tested :**

1. an ingress policy; exposes the workload on ingress to entire-cluster on named ports; since the dst is the workload itself - converts the ports to numbers

ing_spec_result
 
2. an egress policy - which exposes the workload to entire cluster ; on named ports : we will see these potential named ports on the result.

egress_res

3. an example combining named ports and port numbers:

example_spec_res

shireenf-ibm:
next tasks to be done are:
- adding new api func to the policy-engine; so all pre-process (and storing general conns for pods) will be computed only for exposure-analysis.
- in case of exposure-analysis; updating some computation that can benefit from the stored data of the policy

since this PR is already labeled `L` ; I think should commit these to a new PR ?

adisos:
> next tasks to be done are:
> 
>     * adding new api func to the policy-engine; so all pre-process (and storing general conns for pods) will be computed only for exposure-analysis.
> 
>     * in case of exposure-analysis; updating some computation that can benefit from the stored data of the policy
> 
> 
> since this PR is already labeled `L` ; I think should commit these to a new PR ? (sub-task : * - [ ] adding new api func for policy-engine for running exposure-analysis individually)

ok, let's have these in a new PR

pkg/netpol/eval/internal/k8s/netpol.go
adisos: thanks, the example seem to make sense

pkg/netpol/internal/common/portset.go
```diff
@@ -102,18 +110,44 @@ func (p *PortSet) IsAll() bool {
...
 func (p *PortSet) Contains(port int64) bool {
-	portObj := PortSet{}
+	portObj := MakePortSet(false)
 	portObj.AddPortRange(port, port)
 	return portObj.ContainedIn(*p)
 }
 
+// GetNamedPortsKeys returns the named ports of current portSet
+func (p *PortSet) GetNamedPortsKeys() []string {
+	res := []string{}
```
adisos, line 144: pre allocate res based on size of map

pkg/netpol/internal/common/portset.go
```diff
@@ -102,18 +110,44 @@ func (p *PortSet) IsAll() bool {
...
+		// if p.Ports is empty but p.NamedPorts is not: start a new string
+		if res == emptyStr {
+			res = ""
+		} else {
+			res += comma
+		}
+		for _, k := range sortedNamedPorts {
+			res += k + comma
+		}
+		res = res[:len(res)-1]
```
adisos, line 130: can be replaced with `res += strings.Join(sortedNamedPorts, comma)`

pkg/netpol/eval/resources.go
```diff
@@ -311,6 +311,13 @@ func (pe *PolicyEngine) upsertPod(pod *corev1.Pod) error {
 	return nil
 }
 
+func initiatePolicyGeneralConns() k8s.PolicyGeneralRulesConns {
```
adisos, line 314: rename to `initPolicyGeneralConns()`

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -255,31 +255,41 @@ func (pod *Pod) PodExposedTCPConnections() *common.ConnectionSet {
...
 }
 
 // updatePodXgressExposureToEntireClusterData updates the pods' fields which are related to entire class exposure on ingress/egress
-func (pod *Pod) updatePodXgressExposureToEntireClusterData(ruleConns *common.ConnectionSet, isIngress bool) {
+func (pod *Pod) UpdatePodXgressExposureToEntireClusterData(ruleConns *common.ConnectionSet, isIngress bool) {
 	if isIngress {
-		pod.IngressExposureData.EntireClusterConnection.Union(ruleConns)
+		// for a dst pod check if the given ruleConns contains namedPorts; and replace them with pod's
+		// matching port number
+		ruleConnsCopy := ruleConns.Copy() // copying the connectionSet; in order to replace
```
adisos, line 283: no need to copy if the conn has no named ports

pkg/netpol/eval/resources.go
```diff
@@ -320,11 +327,21 @@ func (pe *PolicyEngine) upsertNetworkPolicy(np *netv1.NetworkPolicy) error {
 	if _, ok := pe.netpolsMap[netpolNamespace]; !ok {
 		pe.netpolsMap[netpolNamespace] = make(map[string]*k8s.NetworkPolicy)
 	}
-	pe.netpolsMap[netpolNamespace][np.Name] = (*k8s.NetworkPolicy)(np)
+
+	newNP := &k8s.NetworkPolicy{
```
adisos, line 331: `newNP` can be renamed to `newNetpol`

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -58,15 +76,22 @@ func (np *NetworkPolicy) convertNamedPort(namedPort string, pod *Pod) int32 {
...
+// if input port is a named port, and the dst peer is nil or  does not have a matching named port defined, return
 // an empty range represented by (-1,-1)
-func (np *NetworkPolicy) getPortsRange(port *intstr.IntOrString, endPort *int32, dst Peer) (startNum, endNum int32, err error) {
+func (np *NetworkPolicy) getPortsRange(port *intstr.IntOrString, endPort *int32, dst Peer) (startNum, endNum int32,
+	namedPort string, err error) {
 	var start, end int32
+	portName := ""
 	if port.Type == intstr.String {
+		if dst == nil {
+			return -1, -1, port.StrVal, nil
```
adisos, line 88: replace `-1` with `common.NoPort`

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -94,19 +119,23 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
...
+			startPort, endPort, portName, err := np.getPortsRange(rulePorts[i].Port, rulePorts[i].EndPort, dst)
 			if err != nil {
 				return res, err
 			}
-			if isEmptyPortRange(startPort, endPort) {
-				continue
+			if dst == nil && portName != "" {
+				// adding namedPort to connectionSet in case of :
+				// -  dst is nil - for general connections;
+				// - TODO: if dst is a fake pod (representative)
```
adisos, line 133: why is this a `todo` in the comment?

shireenf-ibm: i wanted to focus on general conns only in this PR (and test it locally);
so put this as TODO for next PR so I'll remeber to update and test

adisos: The question was what is planned as a todo? Why does it require additional handling here?

shireenf-ibm: ah;
it serves the case of : a connection with named port when the dst peer is "representative";
in this case we can not convert the named port to a number; but still want to keep the "named port" in the connection as the potential port;

adisos: but it seems to already be handled by the current code? how is it different from a real pod that does not have the specified named port?

shireenf-ibm: when a connection has "namedPort"; but the real pod does not actually has the named-port in its spec; this specified port/connection will be ignored and not saved as allowed connection between the peers.
if the real pod has the named-port in its spec; the number of the port will be appended to the allowed connection.

but in case of representative pod; the named-port may be a potential port for the connection

adisos (approved):

PR #307, opened 2024-02-12
Policy engine with new api func for exposure analysis 

shireenf-ibm:
issue #236

sub task : 

>   - [ ] adding new api func for policy-engine for running exposure-analysis individually([info](https://github.com/np-guard/netpol-analyzer/pull/306#issuecomment-1933431823))

PR's tasks :

- [x] adding new api func to the policy-engine; so all pre-process (and storing general conns for pods) will be computed only for exposure-analysis. (https://github.com/np-guard/netpol-analyzer/pull/307/commits/39172376d0e9867c2b42b16d8f41b8347872728b)

- [x] in case of exposure-analysis; updating some computation that can benefit from the stored data of the policy (https://github.com/np-guard/netpol-analyzer/pull/307/commits/a8474e2eeea1cb86c494be3c9892c9e7a183e352) and (https://github.com/np-guard/netpol-analyzer/pull/307/commits/e5a5270070c7bf3ad554df5be18bf8710744e253)

- tested locally with desired results

### DIFF
--- a/pkg/netpol/connlist/connlist.go
+++ b/pkg/netpol/connlist/connlist.go
@@ -203,6 +203,7 @@ func (ca *ConnlistAnalyzer) getPolicyEngine(objectsList []parser.K8sObject) (*ev
 	if err != nil {
 		return nil, err
 	}
+	// TODO: this will be eliminated when adding representative peers while policies upsert
 	// add representative resources
 	err = pe.SetExposureAnalysisResources()
 	return pe, err
--- a/pkg/netpol/connlist/connlist.go
+++ b/pkg/netpol/connlist/connlist.go
@@ -190,18 +190,29 @@ func (ca *ConnlistAnalyzer) hasFatalError() error {
 	return nil
 }
 
-func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8sObject) ([]Peer2PeerConnection, []Peer, error) {
+// getPolicyEngine returns a new policy engine considering the exposure analysis option
+func (ca *ConnlistAnalyzer) getPolicyEngine(objectsList []parser.K8sObject) (*eval.PolicyEngine, error) {
 	// TODO: do we need logger in policyEngine?
-	pe, err := eval.NewPolicyEngineWithObjects(objectsList)
+	if !ca.exposureAnalysis {
+		return eval.NewPolicyEngineWithObjects(objectsList)
+	}
+	// else build new policy engine with exposure analysis
+	pe := eval.NewPolicyEngineWithOptions(ca.exposureAnalysis)
+	// add objects from real resources
+	err := pe.AddObjects(objectsList)
+	if err != nil {
+		return nil, err
+	}
+	// add representative resources
+	err = pe.SetExposureAnalysisResources()
+	return pe, err
+}
+
+func (ca *ConnlistAnalyzer) connslistFromParsedResources(objectsList []parser.K8sObject) ([]Peer2PeerConnection, []Peer, error) {
+	pe, err := ca.getPolicyEngine(objectsList)
 	if err != nil {
 		ca.errors = append(ca.errors, newResourceEvaluationError(err))
 		return nil, nil, err
-	}
-	if ca.exposureAnalysis {
-		err = pe.SetExposureAnalysisResources()
-		if err != nil {
-			return nil, nil, err
-		}
 	}
 	ia, err := ingressanalyzer.NewIngressAnalyzerWithObjects(objectsList, pe, ca.logger, ca.muteErrsAndWarns)
 	if err != nil {
@@ -491,7 +502,7 @@ func (ca *ConnlistAnalyzer) existsFocusWorkload(peers []Peer, excludeIngressAnal
 func (ca *ConnlistAnalyzer) getConnectionsBetweenPeers(pe *eval.PolicyEngine, peers []Peer) ([]Peer2PeerConnection, exposureMap, error) {
 	connsRes := make([]Peer2PeerConnection, 0)
 	exposuresMap := exposureMap{}
-	// sets for marking peer checked for ingress/egress exposure to entire cluster data once
+	// for exposure-analysis use: sets for marking peer checked for ingress/egress exposure to entire cluster data once
 	ingressSet := make(map[Peer]bool, 0)
 	egressSet := make(map[Peer]bool, 0)
 
--- a/pkg/netpol/eval/internal/k8s/netpol.go
+++ b/pkg/netpol/eval/internal/k8s/netpol.go
@@ -294,7 +294,7 @@ func (np *NetworkPolicy) EgressAllowedConn(dst Peer, protocol, port string) (boo
 	return false, nil
 }
 
-// GetEgressAllowedConns returns the set of allowed connetions from any captured pod to the destination peer
+// GetEgressAllowedConns returns the set of allowed connections from any captured pod to the destination peer
 // if exposureFlag is true; the result is initiated with the matching policy's general connections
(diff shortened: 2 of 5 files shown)


pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -289,12 +295,20 @@ func (np *NetworkPolicy) EgressAllowedConn(dst Peer, protocol, port string) (boo
 }
 
 // GetEgressAllowedConns returns the set of allowed connetions from any captured pod to the destination peer
```
adisos, line 297:
```suggestion
// GetEgressAllowedConns returns the set of allowed connections from any captured pod to the destination peer
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -289,12 +295,20 @@ func (np *NetworkPolicy) EgressAllowedConn(dst Peer, protocol, port string) (boo
...
-func (np *NetworkPolicy) GetEgressAllowedConns(dst Peer) (*common.ConnectionSet, error) {
+// if exposureFlag is true; the result is initiated with the matching policy's general connections
+// and computations of rule conns are skipped for general rules;
+func (np *NetworkPolicy) GetEgressAllowedConns(dst Peer, exposureFlag bool) (*common.ConnectionSet, error) {
 	res := common.MakeConnectionSet(false)
+	if exposureFlag {
+		// if exposure-analysis is on; we already have the policy's entire world and entire cluster connections;
+		// so we can initiate the result with it
+		res = np.initEgressAllowedConnsFromPolicyGeneralConns(dst)
```
adisos, line 305: I don't think this initialization has much value for being more efficient here... as the current implementation iterates the rules anyway... I don't see how it helps?

I thought we could avoid iterating the rules, in cases where we know the policy allows all connections for all targets (either cluster-internal or also external).

pkg/netpol/connlist/connlist.go
```diff
@@ -190,19 +190,30 @@ func (ca *ConnlistAnalyzer) hasFatalError() error {
...
+	}
+	// else build new policy engine with exposure analysis
+	pe := eval.NewPolicyEngineWithOptions(ca.exposureAnalysis)
+	// add objects from real resources
+	err := pe.AddObjects(objectsList)
+	if err != nil {
+		return nil, err
+	}
+	// add representative resources
+	err = pe.SetExposureAnalysisResources()
```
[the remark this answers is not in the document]
shireenf-ibm: yes we do : 

> * - [ ]  optimize fake-pods generations 
        - first add `fake pods` for all non-empty rules while policies upsert
        - refine pods that has a match in the resources

added TODO

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -289,12 +295,20 @@ func (np *NetworkPolicy) EgressAllowedConn(dst Peer, protocol, port string) (boo
 }
 
 // GetEgressAllowedConns returns the set of allowed connetions from any captured pod to the destination peer
-func (np *NetworkPolicy) GetEgressAllowedConns(dst Peer) (*common.ConnectionSet, error) {
+// if exposureFlag is true; the result is initiated with the matching policy's general connections
+// and computations of rule conns are skipped for general rules;
```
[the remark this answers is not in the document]
shireenf-ibm: will be reverted if we don't want to init the connections with general

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: I see; yes the initialization may just save the computation of rule connections `np.ruleConnections(rulePorts, dst)` for the general rules (as it was already stored while pre-proccessing); 
1. so should revert the commits of `in case of exposure-analysis; updating some computation that can benefit from the stored data of the policy (https://github.com/np-guard/netpol-analyzer/commit/a8474e2eeea1cb86c494be3c9892c9e7a183e352) and (https://github.com/np-guard/netpol-analyzer/commit/e5a5270070c7bf3ad554df5be18bf8710744e253)` ?

i think the only policies that we can avoid iterating their rules are the policies with "general" rules only;
 a specified rule might always has "extra" value to a specific connection between two peers.
 
 2. if desired; we can while pre-processing store if the policy contains only the general rule/s? (don't know if it is very common use-case)

adisos: what do you mean by policies with "general" rules only?

> a specified rule might always has "extra" value to a specific connection between two peers

how if we avoid iterating only for "all connections" allowed for all targets ?

adisos: > the initialization may just save the computation of rule connections 

This does not save much, so I would remove that to simplify the code

shireenf-ibm: > what do you mean by policies with "general" rules only?
>
I mean policies that the only rules that they contain are either "empty from/to " - which means all targets
or it only from/to rules are with : `namespaceSelector: {}`; i.e. the policies that their conns are stored for sure
 
> > a specified rule might always has "extra" value to a specific connection between two peers
> 
> how if we avoid iterating only for "all connections" allowed for all targets ?

we can avoid iterating policies above (in this comment) and not only all connections;
 we might have a policy for all destinations with specific ports

shireenf-ibm: > > the initialization may just save the computation of rule connections
> 
> This does not save much, so I would remove that to simplify the code

reverted

adisos: > we can avoid iterating policies above (in this comment) and not only for all connections;
we might have a policy for all destinations with specific ports

But this requires checking in advance containment of all rules' connections.. maybe add a comment about this but avoid this more complex approach now

shireenf-ibm: I think we have 2 cases: 
1- policies that include only "general" rules, for example:

```
egress:
  - to:
    - namespaceSelector: {}
    ports:
        - port: 8000
 ingress:
    - ports:
           - port: 8090
// which is egress to entire cluster on TCP 8000
// and ingress from whole world on TCP 8090
```

for this policy; both "general" connections are stored on pre-process
I can save also a new variable for the policy that indicates that there are only general (no other) rules in the policy;
and then avoid iterating it when computing connections between two peers 
but is this a common use-case? should apply this?

2. when the policy has mix of general and more specified rules (with non-empty namespaceSelector/ podSelector)
in this case we need to check rules' connections containment in order to avoid iterating later (complex approach)

pkg/netpol/eval/internal/k8s/netpol.go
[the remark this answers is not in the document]
shireenf-ibm: not relevant any more

pkg/netpol/eval/internal/k8s/netpol.go
adisos: I don't see why we need to differentiate between these cases.. if we have mixed rules, we can avoid iterating the rules if we identified in the pre-prcoessing that there was a rule enabling all connections to the entire cluster..

shireenf-ibm: uh, yes when the "general" connection is "All Connections" ; there is no need to iterate; (committed changes)

I was speaking on the case where the general stored connections are not "All Connections" .. trying to think how to benefit in such cases from stored data (however this will complicate the code, if desired may consider in another issue)

shireenf-ibm: done

adisos (approved):

PR #314, opened 2024-02-28
Optimize representative peers generation 

shireenf-ibm:
#236 

sub-task :

> * - [ ]  optimize fake-pods generations 
        - first add `fake pods` for all non-empty rules while policies upsert
        - refine pods that has a match in the resources

in this PR:
1. created a representative peer for each rule with only non-empty `namespaceSelector` in the policies
2. refined while upserting pods and workloads (policies upserted first)

### DIFF
--- a/pkg/netpol/eval/exposure.go
+++ b/pkg/netpol/eval/exposure.go
@@ -95,3 +95,63 @@ func (pe *PolicyEngine) GetPeerNsLabels(p Peer) (map[string]string, error) {
 	}
 	return peer.PotentialNamespaceLabels, nil
 }
+
+// IsRedundantRepresentativePeer returns wether the given peer is a redundant representative peer or not
+func (pe *PolicyEngine) IsRedundantRepresentativePeer(peer Peer) bool {
+	if !pe.IsRepresentativePeer(peer) {
+		return false
+	}
+	return peer.(*k8s.RepresentativePeer).IsRedundant
+}
+
+/////////
+
+// RepresentativePeerMatchesRealWorkloadPeer gets two peers
+// if one is a WorkloadPeer and the other is a RepresentativePeer: returns whether the namespaceLabels (TODO: and podLabels)
+// of the RepresentativePeer are contained in the labels of the WorkloadPeer's namespace
+func (pe *PolicyEngine) RepresentativePeerMatchesRealWorkloadPeer(firstPeer, secondPeer Peer) bool {
+	var repPeer *k8s.RepresentativePeer
+	var wlPeer *k8s.WorkloadPeer
+	if pe.IsRepresentativePeer(firstPeer) {
+		repPeer = firstPeer.(*k8s.RepresentativePeer)
+		// if the second is not a workload peer - no match for sure ; return false
+		if wlPeer, _ = isPeerAWorkloadPeer(secondPeer); wlPeer == nil {
+			return false
+		}
+		// TODO : compare also podLabels
+		return pe.potentialNamespaceLabelsContainedInRealLabels(repPeer, wlPeer)
+	} // else
+	wlPeer, _ = isPeerAWorkloadPeer(firstPeer)
+	if !pe.IsRepresentativePeer(secondPeer) || wlPeer == nil {
+		return false
+	}
+	repPeer = secondPeer.(*k8s.RepresentativePeer)
+	return pe.potentialNamespaceLabelsContainedInRealLabels(repPeer, wlPeer)
+}
+
+// potentialNamespaceLabelsContainedInRealLabels checks if the potential namespaceLabels of the fake peer's namespace are contained in
+// the labels of the namespace of the real workloadpeer;
+// if yes, mark the representative peer as redundant
+func (pe *PolicyEngine) potentialNamespaceLabelsContainedInRealLabels(repPeer *k8s.RepresentativePeer, wlPeer *k8s.WorkloadPeer) bool {
+	// todo : special case: add check for namespaceName and the "name" label of fake namespace
+	potentialNsLabels := repPeer.PotentialNamespaceLabels
+	realPodPeer, _ := pe.convertPeerToPodPeer(wlPeer) // todo: handle err
+	actualWlNsLabels := realPodPeer.NamespaceObject.Labels
+	if isMapContainedInOther(potentialNsLabels, actualWlNsLabels) {
+		repPeer.IsRedundant = true
+		return true
+	}
+	return false
+}
+
+func isMapContainedInOther(subMap, other map[string]string) bool {
+	if len(subMap) > len(other) {
+		return false
+	}
+	for k, v := range subMap {
+		if otherV, ok := other[k]; !ok || v != otherV {
+			return false
+		}
+	}
+	return true
+}
--- a/pkg/netpol/eval/exposure.go
+++ b/pkg/netpol/eval/exposure.go
@@ -16,6 +16,10 @@ package eval
 import (
 	"errors"
 	"fmt"
+	"sort"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/np-guard/netpol-analyzer/pkg/internal/netpolerrors"
 	"github.com/np-guard/netpol-analyzer/pkg/netpol/eval/internal/k8s"
@@ -27,10 +31,12 @@ import (
 // any fake namespace added will start with following prefix for ns name and following pod name
 const repNsNamePrefix = "representative-namespace-"
 
-// generateRepresentativePeers adds representative pods with namespace or pod labels inferred from given selectors
-// which are extracted from network policies rules.
-// for example, if a rule within policy has namespace selector "name: foo", then a representative pod in such a
-// namespace with those labels will be added, representing all potential pods in such a namespace
+// generateRepresentativePeers : generates and adds to policy engine representative peers where each peer
+// has namespace and pod labels inferred from single policy rule labels in the given list of selectors;
+// for example, if a rule within policy has namespaceSelector "foo: managed", then a representative pod in such a
+// namespace with those labels will be added, representing all potential pods in such a namespace.
+// generated representative peers are unique; i.e. if different rules (e.g in different policies or different directions) has same labels :
+// one representative peer is generated to represent both
 func (pe *PolicyEngine) generateRepresentativePeers(selectorsLabels []k8s.SingleRuleLabels, policyName string) error {
 	for i := range selectorsLabels {
 		_, err := pe.AddPodByNameAndNamespace(k8s.RepresentativePodName, repNsNamePrefix+policyName+fmt.Sprint(i), &selectorsLabels[i])
@@ -39,6 +45,36 @@ func (pe *PolicyEngine) generateRepresentativePeers(selectorsLabels []k8s.Single
 		}
 	}
 	return nil
+}
+
+const comma = ","
+
+// getSortedLabelsString returns a sorted string of the given labels  - helping func
+// @todo on podSelector PR : add also sorted podSelector labels to the returned value (for the map key)
+func getSortedLabelsString(selectorLabels *k8s.SingleRuleLabels) string {
+	nsSelectorStr := labels.SelectorFromSet(labels.Set(selectorLabels.NsLabels)).String()
+	selectorSlice := strings.Split(nsSelectorStr, comma)
+	sort.Strings(selectorSlice)
+	return strings.Join(selectorSlice, comma)
+}
+
+// refineRepresentativePeersMatchingLabels removes from the policy engine all representative peers
+// with labels matching the given labels of a real pod
+func (pe *PolicyEngine) refineRepresentativePeersMatchingLabels(realPodLabels, realNsLabels map[string]string) {
+	keysToDelete := make([]string, 0)
+	// look for representative peers with labels matching the given real pod's (and its namespace) labels
+	for key, peer := range pe.representativePeersMap {
+		potentialPodSelector := labels.SelectorFromSet(labels.Set(peer.Pod.Labels))
+		potentialNsSelector := labels.SelectorFromSet(labels.Set(peer.PotentialNamespaceLabels))
+		if potentialPodSelector.Matches(labels.Set(realPodLabels)) && potentialNsSelector.Matches(labels.Set(realNsLabels)) {
+			keysToDelete = append(keysToDelete, key)
+		}
+	}
+
+	// delete redundant representative peers
+	for _, k := range keysToDelete {
+		delete(pe.representativePeersMap, k)
+	}
 }
 
 // /////////////////////////////////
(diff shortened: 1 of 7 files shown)


adisos:
> peers in the exposureMap to be excluded from the final result []ExposedPeer

why is this related to the next PR? it cannot be done in this PR? 

> to handle the case of two different representative peers with same labels 

what does this mean? (what was the planned handling?)
it should not be possible to generate multiple representative peers with same labels.. consider add caching by labels hash in this PR?

pkg/netpol/eval/resources.go
```diff
@@ -352,11 +344,16 @@ func (pe *PolicyEngine) upsertNetworkPolicy(np *netv1.NetworkPolicy) error {
...
+	// for exposure analysis only: scan policy ingress and egress rules:
+	// 1. to store allowed connections to entire cluster and to all destinations (if such connections are allowed by the policy)
+	// 2. to get selectors and generate representativePeers (each specified rule, gets a representative peer)
 	if pe.exposureAnalysisFlag {
-		err = newNetpol.DetermineGeneralConnectionsOfPolicy()
+		rulesSelectors, scanErr := newNetpol.ScanPolicyRulesForGeneralConnsAndRepresentativePeers()
+		if scanErr != nil {
+			return scanErr
+		}
+		err = pe.generateRepresentativePeers(rulesSelectors, np.Name)
```
adisos, line 356: how about consider to sort the objects to add, so that netpols are added first, and then when workloads are add - for each added workload we can remove a redundant representative peer?
this may simplify the code, and let us avoid having redundant peers through the entire flow?

shireenf-ibm:
> > peers in the exposureMap to be excluded from the final result []ExposedPeer
> 
> why is this related to the next PR? it cannot be done in this PR?
> 

with current "pattern" of exposureMap - (`*peerExposureData`) it can not be done in this PR.
currently the `*peerExposureData` has xgressExposureData as list where each item of it includes `namespaceLabels ` (which are inferred from one representativePeer )
in next task (* - [ ] optimize exposure connections refinement in connlist pkg ); the plan is to change the internal exposure data struct to have a map from representativePeer (which include its potentialNsLabels) to the potential connectivity;
so then we can refine redundant representativePeers.
(so in next task (next optimize PR) the `XgressExposureData` interface might be changed to have map from PotentialLabels to PotentialConnectivity and the `ExposedPeer` interface will be changed from having  lists of `[]XgressExposureData` to single `XgressExposureData`

pkg/netpol/eval/resources.go
shireenf-ibm: actually I can sort only in case of exposure-analysis

adisos: ok, so let's try to simplify.
for any added workload you will have to iterate the remaining representative peers and remove those which are of a matching selector, right?

pkg/netpol/eval/exposure.go
```diff
@@ -147,3 +95,97 @@ func (pe *PolicyEngine) GetPeerNsLabels(p Peer) (map[string]string, error) {
...
+	if nameLabelKey != "" && realPodPeer.NamespaceObject.Name != potentialNsLabels[nameLabelKey] {
+		return false, nil
+	}
+	isRedundant := isMapContainedInOther(potentialNsLabels, actualWlNsLabels, nameLabelKey)
+	repPeer.IsRedundant = isRedundant
+	return isRedundant, nil
+}
+
+// isMapContainedInOther checks if the subMap is contained in the other map;
+// ignored key : is the name key; its value is already compared so to be ignored if not empty
```
adisos, line 156: why is this key ignored? 

+ cannot use existing methods for checking selectors matching ?

shireenf-ibm: - the name is compared earlier with the actual name of the namespace. 
so skipped (ignored) checking if it is contained in the real labels. 
(i.e. if a representative peer has selector `name: hello-world`; and we have real namespace  with `hello-world` name; we might not see this label in the real namespace; but we want to compare other labels to see if should remove the representative peer or not)
- at this point I had the selectors as maps not Selector types. (so couldn't use checking selectors matching funcs)

adisos: maybe it is possible to convert maps to selector types? or create selectors from maps?

shireenf-ibm: both are acceptable;
for example: 
if we have a rule in the policy : 
```
from:
    - namespaceSelector:       
       matchLabels:
          kubernetes.io/metadata.name: payments
```

it will match any namespace with this name (even if the namespace dosen't has a label in its labels list with same  `kubernetes.io/metadata.name: payments`.

shireenf-ibm: so while comparing potentialNamespceLabels of representativePeer and a real workload
I first check if the representative peer has this "name" label in its `potentialNamespceLabels ` ; if yes I comapre the value of the name key with the real workload's namespace name:
- if they are different then I return that there is no match between the peers
- if they are the same; I continue to compare the other labels in `potentialNamespceLabels ` are contained in the real namespace labels

adisos: but we add the label `kubernetes.io/metadata.name` to any namespace that we analyze...

shireenf-ibm: oops, forgot about that

pkg/netpol/eval/resources.go
shireenf-ibm: done

pkg/netpol/eval/exposure.go
shireenf-ibm: this func was deleted - handling comparison with real peers code was changed

shireenf-ibm:
> > peers in the exposureMap to be excluded from the final result []ExposedPeer
> 
> why is this related to the next PR? it cannot be done in this PR?
> 
> > to handle the case of two different representative peers with same labels
> 
> what does this mean? (what was the planned handling?) it should not be possible to generate multiple representative peers with same labels.. consider add caching by labels hash in this PR?

no further todo's like those related to this PR

pkg/netpol/eval/resources.go
```diff
@@ -543,8 +600,15 @@ func (pe *PolicyEngine) AddPodByNameAndNamespace(name, ns string, nsLabels map[s
 		return nil, err
 	}
 	if pe.exposureAnalysisFlag && newPod.Name == k8s.RepresentativePodName { // if exposure-analysis and this is not a fake ingress-controller
+		// first compute a selector sorted (and unique) string to be used as a map key
+		// todo : the key of the map might also be the representativePeer.String() if fits (sorted and ensures equality)
+		// (when we decide how to implement it)?!
+		keyStrFromLabels := getSortedLabelsString(objLabels)
```
adisos, line 606: The computation of  unique key from set  of labels already exists in a simpler version in `variantFromLabelsMap()` within `pkg\netpol\eval\internal\k8s\pod.go`.
Can remove implementation of `getSortedLabelsString`, and instead use this mechanism here as well.

pkg/netpol/eval/exposure.go
```diff
@@ -26,71 +28,53 @@ import (
...
-		selectorMap, err := metav1.LabelSelectorAsMap(selector)
+// generateRepresentativePeers : generates and adds to policy engine representative peers where each peer
+// has namespace and pod labels inferred from single policy rule labels in the given list of selectors;
+// for example, if a rule within policy has namespaceSelector "foo: managed", then a representative pod in such a
+// namespace with those labels will be added, representing all potential pods in such a namespace.
+// generated representative peers are unique; i.e. if different rules (e.g in different policies or different directions) has same labels :
+// one representative peer is generated to represent both
+func (pe *PolicyEngine) generateRepresentativePeers(selectorsLabels []k8s.SingleRuleLabels, policyName string) error {
+	for i := range selectorsLabels {
+		_, err := pe.AddPodByNameAndNamespace(k8s.RepresentativePodName, repNsNamePrefix+policyName+fmt.Sprint(i), &selectorsLabels[i])
```
adisos, line 42: are all representative pods going to have the same name?
could have a case of 2 such pods in the same namespace but with different pod selectors? 

also, move to another function the code that returns a namespace string for new representative peer.

pkg/netpol/eval/resources.go
```diff
@@ -531,8 +583,13 @@ func (pe *PolicyEngine) ConvertPeerNamedPort(namedPort string, peer Peer) (int32
...
 // AddPodByNameAndNamespace adds a new fake pod to:
 // the pe.podsMap in case of fake ingress-controller pods
-// or the pe.representativePeersList in case of exposure-analysis peers
-func (pe *PolicyEngine) AddPodByNameAndNamespace(name, ns string, nsLabels map[string]string) (Peer, error) {
+// or the pe.representativePeersMap in case of exposure-analysis peers
+func (pe *PolicyEngine) AddPodByNameAndNamespace(name, ns string, objLabels *k8s.SingleRuleLabels) (Peer, error) {
+	var nsLabels map[string]string
+	if objLabels != nil {
+		nsLabels = objLabels.NsLabels
+	}
```
adisos, line 591: The map can be different from `nil` but with `len` that equals 0.. thus better check comparison of len to 0 instead of comparison to `nil` (if it's nil then also the len is 0).

Should not continue with this function if the len of input labels is 0?

pkg/netpol/eval/resources.go
```diff
@@ -52,26 +53,67 @@ func NewPolicyEngine() *PolicyEngine {
 		podOwnersToRepresentativePodMap: make(map[string]map[string]*k8s.Pod),
 		cache:                           newEvalCache(),
 		exposureAnalysisFlag:            false,
+		representativePeersMap:          nil, // initialized only for exposure analysis
```
adisos, line 56:
```suggestion
```

pkg/netpol/eval/resources.go
```diff
@@ -52,26 +53,67 @@ func NewPolicyEngine() *PolicyEngine {
...
+		err = pe.UpsertObject(policyObj.Networkpolicy)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// AddObjectsByKind adds different k8s objects from parsed resources to the policy engine
+func (pe *PolicyEngine) AddObjectsByKind(objects []parser.K8sObject) error {
```
adisos, line 116: why is this function exported?

pkg/netpol/eval/resources.go
```diff
@@ -52,26 +53,67 @@ func NewPolicyEngine() *PolicyEngine {
...
+// TBD: may be deprcated and use only AddObjectsByKind even when kind is always Networkpolicy?
+func (pe *PolicyEngine) addPolicies(policies []parser.K8sObject) (err error) {
+	for _, policyObj := range policies {
+		err = pe.UpsertObject(policyObj.Networkpolicy)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
adisos, line 113: this function is indeed not required, can just use `AddObjectsByKind`

pkg/netpol/eval/resources.go
```diff
@@ -52,26 +53,67 @@ func NewPolicyEngine() *PolicyEngine {
...
 func (pe *PolicyEngine) AddObjects(objects []parser.K8sObject) error {
+	if !pe.exposureAnalysisFlag { // should not be true ever
+		return nil
+	}
+	policies, nonPolicies := splitPoliciesAndOtherObjects(objects)
+	err := pe.addPolicies(policies)
+	if err != nil {
+		return err
+	}
+	err = pe.AddObjectsByKind(nonPolicies)
```
adisos, line 88: why splitting and having two separate calls?
can just sort the original objects list, using a comparator function, and have the original call on the sorted list of objects.

pkg/netpol/eval/internal/k8s/namespace.go
```diff
@@ -39,7 +39,8 @@ func NamespaceFromCoreObject(ns *corev1.Namespace) (*Namespace, error) {
 	for k, v := range ns.Labels {
 		n.Labels[k] = v
 	}
-
+	// @todo/tbd : should also add the name label as "name:<val>"  or assume policy rules
```
adisos, line 42: I think that `name` key is not auto generated, as opposed to `kubernetes.io/metadata.name`.. please verify. If that is the case, should be able to rely on the labels from the namespace resource itself?

adisos (changes requested):

pkg/netpol/eval/exposure.go
shireenf-ibm: in this case (supporting only ns-selector ) only one pod will be created in each new ns; as each different rule have different namespace labels
so focused only on differentiating the namespace name 

kept "differentiating" the podName for the PR which will support podSelector (supporting both podSelector and nsSelector may require some changes to the for loop)

adisos: ok, please add a todo comment that this will be changed when supporting pod label selector as well

pkg/netpol/eval/resources.go
shireenf-ibm: `objLabels` is not a map; it is a pointer to struct which eventually will include t2o maps (one for namespace labels an the other for pod labels) ; currently it may include only namespace labels 

so I check if this pointer is not nil; then assign `nsLabels` which is declared in line 588 (one line before the if statement) with the struct's `NsLabels`; otherwise we are creating a new pod without any namespace labels (like ingress-controller)

pkg/netpol/eval/resources.go
shireenf-ibm: i thought of implementing a comparator by Kind ; I still will have to till the comparator to put the Networkpolicy objects at first ( and other objects; which I don't care for their order at last); 
seemed to me not more useful than just splitting?

adisos: ok

pkg/netpol/eval/resources.go
adisos: ok

pkg/netpol/eval/internal/k8s/namespace.go
shireenf-ibm: true 
according to [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#automatic-labelling) ;  only `kubernetes.io/metadata.name` is automatically generated for namespaces.

and according to [this](https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetes-io-metadata-name) it is recommended to use `kubernetes.io/metadata.name` as the label selector of name;

so if we have a just `name` key in the rule-selector; we'll need to compare with the actual ns name ?

adisos: > so if we have a just name key in the rule-selector; we'll need to compare with the actual ns name

no, should assume that such key should be defined in the labels of the namespace resource

pkg/netpol/eval/exposure.go

pkg/netpol/eval/resources.go
shireenf-ibm: fixed

pkg/netpol/eval/resources.go

adisos (approved):

PR #316, opened 2024-03-04
exposure analysis test

shireenf-ibm:
issue #236 

sub task : 

>   *  - [ ] basic tests for exposure analysis results


pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,192 @@
...
+			peerList = append(peerList, representativePeers...)
+			peers := convertEvalPeersToConnlistPeer(peerList)
+			ca := NewConnlistAnalyzer(WithExposureAnalysis())
+			_, exposureMap, err := ca.getConnectionsBetweenPeers(pe, peers)
+			require.Nil(t, err, "test %q: error getting connections between peers", tt.testName)
+			exposedPeers := buildExposedPeerListFromExposureMap(exposureMap)
+			require.Equal(t, tt.expectedLenOfExposedPeerList, len(exposedPeers),
+				"test %q: mismatch in length of exposedPeer list", tt.testName)
+			for _, ep := range exposedPeers {
+				require.Contains(t, []string{wl1Name, wl2Name}, ep.ExposedPeer().Name(), "test: %q, unexpected exposed peer name %q",
```
adisos, line 157: missing tests where some of the workloads are not exposed?

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+func checkExpectedVsActualData(t *testing.T, testName string, actualExp ExposedPeer, expectedData expectedPeerResultInfo) {
+	require.Equal(t, expectedData.isEgressProtected, actualExp.IsProtectedByEgressNetpols(),
+		"test: %q, mismatch in is egress protected for peer %q", testName, actualExp.ExposedPeer().String())
+	require.Equal(t, expectedData.isIngressProtected, actualExp.IsProtectedByIngressNetpols(),
+		"test: %q, mismatch in is ingress protected for peer %q", testName, actualExp.ExposedPeer().String())
+	require.Equal(t, expectedData.lenIngressExposedConns, len(actualExp.IngressExposure()),
+		"test: %q, mismatch in length of ingress exposure slice for peer %q", testName, actualExp.ExposedPeer().String())
+	require.Equal(t, expectedData.lenEgressExposedConns, len(actualExp.EgressExposure()),
+		"test: %q, mismatch in length of egress exposure slice for peer %q", testName, actualExp.ExposedPeer().String())
```
adisos, line 191: missing comparison of connection set and namespace labels?

tests/exposure_analysis_tests/test_allow_all/netpol.yaml
```diff
@@ -0,0 +1,16 @@
...
+  podSelector:
+    matchLabels:
+      app: a-app
+  policyTypes:
+  - Ingress
+  - Egress
+  ingress:
+  - {}
+  egress:
+  - {}
```
adisos, line 16: could be more interesting if one side blocks and the other not , such as:
```
  policyTypes:
  - Ingress
  - Egress
  egress:
  - {}
```

shireenf-ibm: - added, all these tests and more tests: will be added for output testing too.
- fixes to exposureMap (from my local "textual output branch" were committed to here); fixing adding the correct `protected data` of a relevant peer  /its new entry in the map

pkg/netpol/connlist/exposure_analysis_test.go
shireenf-ibm: done

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,192 @@
...
+			// get real peers
+			peerList, err := pe.GetPeersList()
+			require.Nil(t, err, "test %q: error getting peer list", tt.testName)
+			representativePeers := pe.GetRepresentativePeersList()
+			require.Equal(t, tt.expectedNumRepresentativePeers, len(representativePeers),
+				"test %q: mismatch in number of representative peers", tt.testName)
+			peerList = append(peerList, representativePeers...)
+			peers := convertEvalPeersToConnlistPeer(peerList)
+			ca := NewConnlistAnalyzer(WithExposureAnalysis())
+			_, exposureMap, err := ca.getConnectionsBetweenPeers(pe, peers)
```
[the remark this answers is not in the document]
adisos: not sure I follow.. the functions from the main flow of `connlist` cannot be used for testing? 
why do we need dedicated code here, that "kind of " repeats the same flow from the main API?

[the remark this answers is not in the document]
shireenf-ibm: ok now I got the question, the main API funcs can be used but I thought that since this test tests only "exposure-analysis" functionality, I should use the relevant funcs only.
will replace the calls with `ca` calls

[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,233 @@
...
+			// workload 2 is not protected at all (unsecure exposed)
+			wl2ExpDataInfo: expectedPeerResultInfo{
+				isIngressProtected:     false,
+				isEgressProtected:      false,
+				lenIngressExposedConns: 0,
+				lenEgressExposedConns:  0,
+			},
+		},
+		{
+			testName:                       "test_multiple_unmatched_rules",
```
adisos, line 148:
```suggestion
			testName:                       "test_multiple_unmatched_rules", // only workload-a in manifests
```

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,233 @@
...
+				ingressExp: []*xgressExposure{
+					newExpDataWithLabelAndTCPConn("foo.com/managed-state", "managed", 8050),
+					newExpDataWithLabelAndTCPConn("release", "stable", -1),
+					newExpDataWithLabelAndTCPConn("effect", "NoSchedule", 8050),
+				},
+				lenEgressExposedConns: 0,
+			},
+		},
+		{
+			testName:                       "test_with_no_netpols",
```
adisos, line 165:
```suggestion
			testName:                       "test_with_no_netpols",  // only workload-a in manifests
```

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,233 @@
...
+			// workload 1 is not protected by any netpol
+			wl1ExpDataInfo: expectedPeerResultInfo{
+				isIngressProtected:     false,
+				isEgressProtected:      false,
+				lenIngressExposedConns: 0,
+				lenEgressExposedConns:  0,
+			},
+		},
+		{
+			testName:                       "test_allow_ingress_deny_egress",
```
adisos, line 177:
```suggestion
			testName:                       "test_allow_ingress_deny_egress", // only workload-a in manifests
```

adisos (changes requested):
looks good, few more comments

adisos (commented):
I find it confusing that the new tests dir for exposure analysis is placed within the the dir of other tests

adisos (commented):
another question: do we handle containment of labels for representative peers? 
for example, a potential connection to ns with label {A: val1} and with conn: all , is including also a potential connection to ns with labels {A: val1, B: val2} , for all contained connections. (which makes such a potential connection redundant).

shireenf-ibm:
> another question: do we handle containment of labels for representative peers? for example, a potential connection to ns with label {A: val1} and with conn: all , is including also a potential connection to ns with labels {A: val1, B: val2} , for all contained connections. (which makes such a potential connection redundant).
> 
> If not, this can be considered in a separate issue.

it is not handled (added a new task )

shireenf-ibm:
> I find it confusing that the new tests dir for exposure analysis is placed within the the dir of other tests
What do you suggest?

the thing is that the `tests` dir is our global testing dir (those tests will be also used in the `connlist_analyzer_test` too for output testing; i.e will be added to the good path tests with the new flag of exposure analysis); so are assumed to be under `tests`. 
I grouped them under a `tests/exposure_analysis_tests` so I can easily remember and find the new tests

pkg/netpol/connlist/exposure_map.go
```diff
@@ -52,11 +47,17 @@ func (ex exposureMap) addPeerUnprotectedData(pe *eval.PolicyEngine, peer Peer, i
...
-	if !protected {
-		if _, ok := ex[peer]; !ok {
-			ex.initiatePeerEntry(peer)
+	_, ok := ex[peer]
+	if !ok && !protected {
+		err = ex.initiateEmptyPeerEntry(pe, peer)
+		return true, err
+	}
+	if ok && protected {
+		if isIngress {
```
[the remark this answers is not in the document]
shireenf-ibm: func `addPeerUnprotectedData` checks if the peer is protected on the given direction:
1- (!ok && !protected) : if it is not protected and it is not in the map already : we add a new empty entry (which means with false protected flags for both directions)- if we have at least one direction which the peer is not protected on; this peer should be in the map (an exposure-analysis case).
2- (ok && protected) if it is protected and there is an entry in the map : so we will update the flag of the current direction as protected 
(might be a case that the peer is not protected on ingress - so an empty entry was created first but called now with egress and got protected - so will update the "egress" unprotected flag to protected )

3. (protected && !ok ) : if it is not in the map and we get protected, we will skip at this stage (cause it might be protected and exposed to only known end-points - which means not exposure-analysis case )

4. (ok && !protected) if ok and not protected: its already in the map with the right data (no need to update ): 
the `ok && protected`  can be replaced with just `ok` (as 4 doesn't change its data)

pkg/netpol/connlist/exposure_map.go
```diff
@@ -86,6 +90,28 @@ func (ex exposureMap) addPeerXgressEntireClusterExp(pe *eval.PolicyEngine, peer
 	return nil
 }
 
+func (ex exposureMap) addNewEntry(pe *eval.PolicyEngine, peer Peer, isEmpty bool) error {
```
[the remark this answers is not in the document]
shireenf-ibm: committed a func docs
 isEmpty indicates if to add an empty entry (with false protected flags) or check the peer's protected data and update accordingly

pkg/netpol/connlist/exposure_map.go
```diff
@@ -10,14 +10,9 @@ import (
...
-// initiatePeerEntry initiates an empty entry for the peer in the exposure map
-func (ex exposureMap) initiatePeerEntry(peer Peer) {
-	ex[peer] = &peerExposureData{
-		isIngressProtected: false,
-		isEgressProtected:  false,
-		ingressExposure:    make([]*xgressExposure, 0),
-		egressExposure:     make([]*xgressExposure, 0),
-	}
+// initiateEmptyPeerEntry initiates an empty entry for the peer in the exposure map
+func (ex exposureMap) initiateEmptyPeerEntry(pe *eval.PolicyEngine, peer Peer) error {
```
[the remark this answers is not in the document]
shireenf-ibm: right, not needed; removed

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
shireenf-ibm: if the "addNewEntry" was called with "isEmpty = false" we need to check again the "protected data" of the peer to be sure that it is correctly updated 

an example for this case:
- when updating the protected data at first we got that the peer is protected on both sides (so an entry will not be added at first stage)
- while checking ingress entire cluster data we found that the peer is exposed, so we will add an entry with peer is exposed to entire cluster on ingress (and automatically update the ingress protected to true); but need to check again for the "egress" protected (the peer might be protected on egress and not exposed to unknown namespaces/entire-cluster) so this func updates its "egress protected" with the correct data.



////////////////////
another solution is to add entries for all peers in the manifests with the protected data (even if its protected and not exposed ) and just refine those protected with no exposure data when building the final result []ExposedPeer

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
adisos: change was not pushed?

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
adisos: > we add a new empty entry (which means with false protected flags for both directions)- if we have at least one direction which the peer is not protected on

so this means that one entry could be initialized incorrectly as `false` ?
maybe the value in this entry should have 3 options: (false/true/unknown)?

pkg/netpol/connlist/exposure_map.go
```diff
@@ -52,11 +42,17 @@ func (ex exposureMap) addPeerUnprotectedData(pe *eval.PolicyEngine, peer Peer, i
 	if err != nil {
 		return false, err
 	}
-	if !protected {
-		if _, ok := ex[peer]; !ok {
-			ex.initiatePeerEntry(peer)
+	_, ok := ex[peer]
+	if !ok && !protected {
```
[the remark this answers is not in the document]
shireenf-ibm: !protected means no netpol at all

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
shireenf-ibm: exactly, will change to be with 3 values

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
shireenf-ibm: protected means there is at least one netpol selecting it but we don't know if it "worth" having an entry in the map (whether the policies selecting the peer expose it insecurely or not)

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
shireenf-ibm: committed changes

tests/exposure_analysis_tests/test_allow_all/netpol.yaml
shireenf-ibm: hi @adisos , committed a new fix of splitting the exposure map into two ingress and egress exposure maps

pkg/netpol/connlist/exposure_analysis.go
```diff
@@ -72,45 +63,72 @@ func xgressExposureListToXgressExposureDataList(xgressExp []*xgressExposure) []X
...
-		if expData.isIngressProtected == protected {
-			ingExp = append(ingExp, expData.ingressExposure...)
+	// first loop the ingressExposureMap : for each peer, fill its ingress exposure data
+	// and check if it exists also in the egressExposureMap too - get its egress exposure data;
+	// otherwise it is protected safely on egress (add default value)
+	for p, ingressExpData := range exposureMaps.ingressExposureMap {
+		// default value for egress exposure
+		egressExposureData := &peerXgressExposureData{
+			isProtected:  true,
+			exposureInfo: nil,
```
adisos, line 105: the code for generating the default object of `peerXgressExposureData` is duplicated (here and in the second loop), can have instead a function such as `func newDefaultXgressExposureData()  *peerXgressExposureData`

pkg/netpol/connlist/exposure_analysis.go
```diff
@@ -4,27 +4,17 @@ import (
...
-// peerExposureData stores exposure data for a peer
-type peerExposureData struct {
-	isIngressProtected int
-	isEgressProtected  int
-	ingressExposure    []*xgressExposure
-	egressExposure     []*xgressExposure
+// peerXgressExposureData store exposure data of a peer on one direction ingress/egress
+type peerXgressExposureData struct {
+	isProtected  bool
+	exposureInfo []*xgressExposure
```
adisos, line 17: can you add documentation on the fields here and of type `xgressExposure` ?
what each field means, and possible relations between fields values. 
for example, if `isProtected` is false, is `exposureInfo` expected to be empty or not .. etc

pkg/netpol/connlist/exposure_map.go
```diff
@@ -35,91 +36,59 @@ func (ex exposureMap) addPeerGeneralExposure(pe *eval.PolicyEngine, peer Peer, i
...
-	if _, ok := ex[peer]; !ok {
-		if isIngress {
-			ex.addNewEntry(peer, protected, unknown)
-		} else {
-			ex.addNewEntry(peer, unknown, protected)
-		}
-	}
-	// update the entry of the peer with the entire cluster connection
+	// exposed to entire cluster - first entry of the peer
+	ex.addNewEntry(peer, true, isIngress)
```
adisos, line 63: why is it certain that the peer is not in the map?

pkg/netpol/connlist/exposure_map.go
```diff
@@ -149,12 +118,14 @@ func (ex exposureMap) addConnToExposureMap(pe *eval.PolicyEngine, allowedConnect
...
-		} else {
-			ex.addNewEntry(peer, unknown, protected)
+	// the peer is protected, check if peer is in the relevant map; if not initialize a new entry
+	if isIngress {
+		if _, ok := ex.ingressExposureMap[peer]; !ok {
+			ex.addNewEntry(peer, true, true)
+		}
+	} else {
+		if _, ok := ex.egressExposureMap[peer]; !ok {
+			ex.addNewEntry(peer, true, false)
```
adisos, line 128: can add the check whether a peer is in the map, to be within `addNewEntry` ?

pkg/netpol/connlist/exposure_map.go
shireenf-ibm: because for each peer we check and update in the relevant map its "general conns" i.e. if unprotected, or there is connection with entire cluster; **only once** and before  adding any other connection with a representative peer.
so since, we have now separate map for each direction , we are for sure know there is no previos entry in the map with this peer. (that was the cause of the "older" fixes in this pr).

however, if I apply [this](https://github.com/np-guard/netpol-analyzer/pull/316#discussion_r1546241978); 
it will always check existence

pkg/netpol/connlist/exposure_analysis.go
shireenf-ibm: done

pkg/netpol/connlist/exposure_analysis.go
shireenf-ibm: done

pkg/netpol/connlist/exposure_analysis.go
```diff
@@ -13,17 +13,24 @@ type exposureMaps struct {
...
+	isProtected bool
+	// exposureInfo list of the exposure-connections data of the peer on the relevant direction.
+	// if isProtected is false (peer is not protected), this field will be empty
 	exposureInfo []*xgressExposure
 }
 
 // ----------------------------------------------------
 // xgressExposure implements XgressExposureData interface
 type xgressExposure struct {
+	// exposedToEntireCluster indicates if the peer is exposed to all namespaces in the cluster for the relevant direction
```
adisos, line 26: if this field is true, then `namespaceLabels` and `podLabels` are empty?

pkg/netpol/connlist/exposure_map.go
```diff
@@ -43,7 +43,7 @@ func (ex *exposureMaps) addPeerUnprotectedData(pe *eval.PolicyEngine, peer Peer,
 		return false, err
 	}
 	if !isProtected {
-		ex.addNewEntry(peer, isProtected, isIngress)
+		ex.checkExistanceAndAddNewEntry(peer, isProtected, isIngress)
```
adisos, line 46: I would keep the name simple, just `addNewEntry`

pkg/netpol/connlist/exposure_map.go
```diff
@@ -71,17 +71,21 @@ func (ex *exposureMaps) addPeerXgressEntireClusterExp(pe *eval.PolicyEngine, pee
 	return nil
 }
 
-// addNewEntry adds a new entry to the relevant ingress/egress map for the given peer
-func (ex *exposureMaps) addNewEntry(peer Peer, isProtected, isIngress bool) {
+// checkExistanceAndAddNewEntry adds a new entry to the relevant ingress/egress map for the given peer
```
adisos, line 74: update comment: `adds a new entry to the relevant ingress/egress map for the given peer, if the peer's key is not in the map`

shireenf-ibm: done

pkg/netpol/connlist/exposure_map.go
shireenf-ibm: done

pkg/netpol/connlist/exposure_analysis.go
shireenf-ibm: yes, updated comments

pkg/netpol/connlist/exposure_analysis_test.go
```diff
@@ -0,0 +1,232 @@
...
+		{
+			testName:                       "test_matched_and_unmatched_rules",
+			expectedNumRepresentativePeers: 1,
+			expectedLenOfExposedPeerList:   2,
+			// workload 1 is protected only on ingress direction and exposed unsecure to entire cluster
+			wl1ExpDataInfo: expectedPeerResultInfo{
+				isIngressProtected:     true,
+				isEgressProtected:      false,
+				lenIngressExposedConns: 1, // connection to unknown dst is contained in entire cluster's conn
+				ingressExp: []*xgressExposure{
```
shireenf-ibm, line 120: this test includes a "connection" containment 
the pod is exposed to any ns with `foo.com/managed-state` on `TCP8050`
and to `entire-cluster` on `TCP8050` : so only the entire cluster conn is in the final result

adisos: ok, good.
do you also have example where both are involved in the final result? (for example the namespace connection is TCP 8050-8051)?

pkg/netpol/connlist/exposure_map.go
```diff
@@ -43,14 +53,15 @@ func (ex *exposureMaps) addPeerUnprotectedData(pe *eval.PolicyEngine, peer Peer,
...
+		ex.addNewEntry(peer, false, isIngress)
 		return true, nil
 	}
 	return false, nil
 }
 
 // addPeerXgressEntireClusterExp checks and adds (if exists) ingress/egress entire cluster exposure for the given peer
-// on the given direction
+// on the given direction;
+// each entry in the relevant map may has at-most one connection with entire-cluster (which is added by this func)
```
adisos, line 64: also relevant to update in documentation for `peerXgressExposureData` ? (that there is at most one such object with entire cluster conn ) within the relevant fiield?

pkg/netpol/connlist/exposure_analysis_test.go
shireenf-ibm: no, added one

pkg/netpol/connlist/exposure_map.go
shireenf-ibm: yes, updated

adisos (approved):
LGTM

PR #333, opened 2024-04-16
dot output with exposure results

shireenf-ibm:
issue #236

task:

> - [ ] adding exposure analysis results in graphical output


pkg/netpol/connlist/conns_formatter_dot.go
```diff
@@ -34,49 +42,139 @@ func peerNameAndColorByType(peer Peer) (nameLabel, color string, isExternal bool
...
+// representative peers and returns the exposure edges and entire cluster line (as external peer line)
+func addExposureOutputData(exposureConns []ExposedPeer, peersVisited map[string]bool,
+	nsPeers, nsRepPeers map[string][]string) (entireClusterLine, exposureEdges []string) {
+	representativeVisited := make(map[string]bool, 0) // acts as a set
+	for _, ep := range exposureConns {
+		if !peersVisited[ep.ExposedPeer().String()] { // an exposed peer is a real peer from the manifests,
+			// updated in the real namespaces map
+			exposedPeerLine, _ := getPeerLine(ep.ExposedPeer())
+			dotformatting.AddPeerToNsGroup(ep.ExposedPeer().Namespace(), exposedPeerLine, nsPeers)
+		}
```
adisos, line 117: how is it possible that a real peer is not visited by this stage? (from the original connections list?)

adisos (commented):
Looks good, added one question.
Another question here: 
Do you have an example with multiple labels per representative namespace? 
for example, `namespace with L1=V1,L2=V2`

pkg/netpol/connlist/conns_formatter_dot.go
shireenf-ibm: possible use-case:  a real peer that is selected by a policy that exposes it to representative peers only on both ingress and egress. (this peer will not appear in the original connlist report)
a test was added (running it twice once without exposure flag and once with , to see that without it does not appear in the report)

adisos (approved):
LGTM

PR #352, opened 2024-05-15
don't remove representative peers in any-namespace

shireenf-ibm:
issue #236 

sub task:

>   * - [ ]  handling cases of empty nsSelector (any-namespace) with non-empty podSelector : don't refine its representative peer if there is a matching pod in a specific namespace

also: keep representative peer  for the case of all pods within a certain namespace


pkg/netpol/eval/exposure.go
```diff
@@ -68,12 +68,19 @@ func (pe *PolicyEngine) extractLabelsAndRefineRepresentativePeers(podObj *k8s.Po
...
 // refineRepresentativePeersMatchingLabels removes from the policy engine all representative peers
 // with labels matching the given labels of a real pod
+// representative peers matching any-namespace will not be removed.
 func (pe *PolicyEngine) refineRepresentativePeersMatchingLabels(realPodLabels, realNsLabels map[string]string) {
 	keysToDelete := make([]string, 0)
 	// look for representative peers with labels matching the given real pod's (and its namespace) labels
 	for key, peer := range pe.representativePeersMap {
 		potentialPodSelector := labels.SelectorFromSet(labels.Set(peer.Pod.Labels))
 		potentialNsSelector := labels.SelectorFromSet(labels.Set(peer.PotentialNamespaceLabels))
+		if potentialNsSelector.Empty() { // representative peer that matches any-namespace, will not be removed
```
adisos, line 78: does it differentiate between
```
namespaceSelector: {}
podSelector: {k: v}
```
(which is any namespace)

to
```
podSelector: {k: v}
```
(which is the network policy namespace)?

adisos (commented):
added one question

pkg/netpol/eval/exposure.go
adisos: what about 
```
namespaceSelector: {k: v}
podSelector: {}
```
this creates exposure to all pods in a certain namespace..
does this require indication as exposure? 
it's not as matching pod with certain labels..

shireenf-ibm: > does it differentiate between
> 
> ```
> namespaceSelector: {}
> podSelector: {k: v}
> ```
> 
> (which is any namespace)
> 
> to
> 
> ```
> podSelector: {k: v}
> ```
> 
> (which is the network policy namespace)?

yes, it does 
examples:
1. `tests/test_conn_with_pod_selector_in_any_ns` with :

> - from:
    - namespaceSelector: {}
      podSelector: 
        matchLabels:
            role: monitoring
            
   (output: `test_outputs\connlist\exposure_test_conn_with_pod_selector_in_any_ns_connlist_output.dot.png`)
            
 2. ` tests/test_conn_with_only_pod_selector`
 
> - from:
    - podSelector: 
        matchLabels:
            role: monitoring          
      (output:`test_outputs\connlist\exposure_test_conn_with_only_pod_selector_connlist_output.dot.png`)

shireenf-ibm: > and what about
> 
> ```
> namespaceSelector: {k: v}
> podSelector: {}
> ```
> 
> this creates exposure to all pods in a certain namespace.. does this require indication as exposure? it's not as matching pod with certain labels..
> 
> do we have any example test capturing this case with current implementation?

we have following example:
test: `tests\test_conn_to_all_pods_in_a_new_ns`
with:

> - from:
    - namespaceSelector: 
        matchLabels:
            kubernetes.io/metadata.name: backend
      podSelector: {}

output: `test_outputs\connlist\exposure_test_conn_to_all_pods_in_a_new_ns_connlist_output.dot.png`

in this example the `backend` ns is not in the manifests and we have no real pods in this ns so we see an exposure to representative peers `backend\[all pods]`.

if we had a real pod in the namespace backend : it would be a match and we won't see the exposure line.(like all cases where the podSelector was not used, if there is a pod in the namespace the it is a match)

adisos: > if we had a real pod in the namespace backend : it would be a match and we won't see the exposure line.

ok, the question is - should we consider changing this behavior? if all pods in a namespace are permitted, it might be relevant as exposure info, even if there is a certain real pod in the namespace. (BTW there could be a real pod in that namespace on which there is a certain policy that blocks such a connection for this pod.. in such a case the connection to this pod is not in the report, nor the exposure to this namespace). [consider adding such an example as well]

adisos: another scenario question: there could be a representative peer with real namespace, in such case there could be network policies for that entire namespace, which are not considered for this representative peer?

shireenf-ibm: > > if we had a real pod in the namespace backend : it would be a match and we won't see the exposure line.
> 
> ok, the question is - should we consider changing this behavior (when the policy permits all pods in a real namespace)? if all pods in a namespace are permitted, it might be relevant as exposure info, even if there is a certain real pod in the namespace.

I see, yes we can change this. 
our assumption was that if there is a rule selecting a namespace and there was a real pod in that namespace so the "developer" knows that connections are allowed to that namespace and didn't add it to the exposure lines.
so by changing this we want to emphasize that conns with all pods in that namespace are allowed . 
( however this will affect also tests with only `namespaceSelector`, where the namespace is real and there is a pod in it in the manifests )


> (BTW there could be a real pod in that namespace on which there is a certain policy that blocks such a connection for this pod.. in such a case the connection to this pod is not in the report, nor the exposure to this namespace). [consider adding such an example as well]

I prefer to add such examples within the ` adding more complicated tests` PR

shireenf-ibm: > another scenario question: there could be a representative peer with real namespace, in such case there could be network policies for that entire namespace, which are not considered for this representative peer?

I am not sure I understood the question.
since k8s network-policies select pods and not namespaces; and if we have a podSelector: {} in a network-policy (means that the policy selects all pods in its namespace)
so if we have a pod in the policy's namespace; all exposure data from this policy will be reported in the exposure of this pod.
but if we don't have any pod in the namespace  of the policy in the manifests so it won't interest us (we can't add exposure data of a non-existing pod)

adisos: I meant to a policy that  selects all pods in its namespace.. (let it be `foo` namespace) 
let's assume we do not have a pod in that namespace
but another pod (`bar/pod1`) has exposure to this namespace (e.g. policy that permits egress to `foo`) 
the question is - do we consider such policy on a representative peer (namespace)? 
(for example, a policy that is a default deny for the `foo` namespace)

adisos: > I think this scenario is for the adding more complicated tests PR too

I think we should add examples and catch unexpected issues ASAP, so I prefer adding more examples we can think of sooner then later.

shireenf-ibm: > I meant to a policy that selects all pods in its namespace.. (let it be `foo` namespace) let's assume we do not have a pod in that namespace but another pod (`bar/pod1`) has exposure to this namespace (e.g. policy that permits egress to `foo`) the question is - do we consider such policy on a representative peer (namespace)? (for example, a policy that is a default deny for the `foo` namespace)

I will add an example to see the behavior and update

shireenf-ibm:
in this [commit](https://github.com/np-guard/netpol-analyzer/pull/352/commits/3c9e43066968b56b7da3aae890bc9908fe00ce12) 
added two examples of exposure to an existing namespace (having real pod in the ns)
with current approach:
- in first test `test_conn_to_all_pods_in_an_existing_ns` since we have a real pod in `backend` we don't see "all pods" exposure line
- in second test `test_conn_to_new_pod_in_an_existing_ns` I used specific `podSelector` and so we see it in the exposure report

pkg/netpol/eval/exposure.go
shireenf-ibm: added an example locally, the exposure analysis of bar still shows the line:
`bar/workload-a[Deployment]      =>      foo/[all pods] : All Connections`
even that we have default-deny-in-foo for ingress 

should fix the bug in this PR?

adisos, line 78:
```suggestion
		if potentialNsSelector.Empty() { 
		// empty --representative peer that matches any-namespace, thus will not be removed
		// note that if the policy had nil namespaceSelector, it would be converted to the namespace of the policy 
```

adisos: create new item and implement in a separate PR .
analysis should generate representative peers by also considering policies that capture them, and separate into sub-sections (sub-peers) that are having different connectivity by the policies (for example pods in ns with certain labels, and pods in ns without these labels).

adisos: > > > if we had a real pod in the namespace backend : it would be a match and we won't see the exposure line.
> > 
> > 
> > ok, the question is - should we consider changing this behavior (when the policy permits all pods in a real namespace)? if all pods in a namespace are permitted, it might be relevant as exposure info, even if there is a certain real pod in the namespace.
> 
> I see, yes we can change this. our assumption was that if there is a rule selecting a namespace and there was a real pod in that namespace so the "developer" knows that connections are allowed to that namespace and didn't add it to the exposure lines. so by changing this we want to emphasize that conns with all pods in that namespace are allowed . ( however this will affect also tests with only `namespaceSelector`, where the namespace is real and there is a pod in it in the manifests )
> 
> > (BTW there could be a real pod in that namespace on which there is a certain policy that blocks such a connection for this pod.. in such a case the connection to this pod is not in the report, nor the exposure to this namespace). [consider adding such an example as well]
> 
> I prefer to add such examples within the ` adding more complicated tests` PR

ok, let's have this change in this PR , including adding the relevant examples , and mark by the example which has inaccurate output, a TODO comment, to make sure it is fixed in the next PR .

pkg/netpol/connlist/connlist_test.go
```diff
@@ -940,4 +940,17 @@ var goodPathTests = []struct {
...
 	},
+	{
+		// following test resources : contains two pods in different namespaces, and two policies, one for each namespace
+		// first policy captures: hello-world/workload-a and exposes it on Ingress to all pods in backend namespace
+		// second policy captures: backend/backend-app and denies all egress from it
+		// so as result hello-world/workload-a is actually exposed to all backend pods except for backend-app
+		// @TODO: following exposure line in output :
+		// `hello-world/workload-a[Deployment]      <=      backend/[all pods] : TCP 8050`
+		// should be replaced with :
+		// `hello-world/workload-a[Deployment]      <=      backend/[all pods except backend/backend-app] : TCP 8050`
```
adisos, line 951:
```
// should be replaced with :
// `hello-world/workload-a[Deployment]      <=      backend/[pods without app: backend-app] : TCP 8050`
```

shireenf-ibm: done

pkg/netpol/connlist/connlist_test.go
```diff
@@ -918,4 +918,39 @@ var goodPathTests = []struct {
...
+	},
+	{
+		// following test resources : contains two pods in different namespaces, and two policies, one for each namespace
+		// first policy captures: hello-world/workload-a and exposes it on Ingress to all pods in backend namespace
+		// second policy captures: backend/backend-app and denies all egress from it
+		// so as result hello-world/workload-a is actually exposed to all backend pods except for backend-app
+		// @TODO: following exposure line in output :
+		// `hello-world/workload-a[Deployment]      <=      backend/[all pods] : TCP 8050`
+		// should be replaced with :
+		// `hello-world/workload-a[Deployment]      <=      backend/[pods without app: backend-app] : TCP 8050`
```
adisos, line 951:
```suggestion
		// note: following exposure line in output :
		// `hello-world/workload-a[Deployment]      <=      backend/[all pods] : TCP 8050`
		// could have been more accurate with:
		// `hello-world/workload-a[Deployment]      <=      backend/[pods without app: backend-app] : TCP 8050`
		// but the goal is to hint where policy can be tightened, thus it is ok to ignore policies that capture 
		// representative peers in the analysis 
		
```
updated comment to a note rather than todo, with explanation of the rational for this case.

adisos (approved):

PR #360, opened 2024-05-22
supporting csv, md and json formats

shireenf-ibm:
issue #236

task:

> - [ ] support all output formats of `connnlist` with exposure analysis


pkg/netpol/connlist/conns_formatter_md.go
```diff
@@ -8,33 +8,59 @@ package connlist
...
-// this format is not supported with exposure analysis; exposureConns is not used;
+// writeOutput returns a md string form of connections from list of Peer2PeerConnection objects,
+// and exposure analysis results from list ExposedPeer if exists
 func (md *formatMD) writeOutput(conns []Peer2PeerConnection, exposureConns []ExposedPeer) (string, error) {
-	mdLines := make([]string, len(conns))
-	for index := range conns {
-		mdLines[index] = getMDLine(formSingleP2PConn(conns[index]))
-	}
-	sort.Strings(mdLines)
+	exposureFlag := len(exposureConns) > 0
```
adisos, line 32: could it be that when running with `--exposure` there is not exposure results, but then the output does not add the header? isn't it a bit misleading output?

test_outputs/connlist/exposure_acs-security-demos_connlist_output.csv
```diff
@@ -0,0 +1,24 @@
...
+{ingress-controller},frontend/webapp[Deployment],TCP 8080
+Exposure Analysis Result:,,
+src,dst,conn
+backend/checkout[Deployment],entire-cluster,UDP 5353
+backend/recommendation[Deployment],entire-cluster,UDP 5353
+backend/reports[Deployment],entire-cluster,UDP 5353
+entire-cluster,frontend/asset-cache[Deployment],TCP 8080
+entire-cluster,frontend/webapp[Deployment],TCP 8080
+frontend/webapp[Deployment],entire-cluster,UDP 5353
+payments/gateway[Deployment],entire-cluster,UDP 5353
```
adisos, line 24: This is a bit inconsistent with `txt` output , where the first item in each line is the actual workload.. 
One option here is to have 2 sections instead of one - ingress exposure and egress exposure.. (that would require also header updates) 
What do you think is better?
@zivnevo what do you think?

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -124,3 +145,63 @@ func getRepresentativePodString(podLabels map[string]string, txtOutFlag bool) st
 	}
 	return res
 }
+
+// following code is common for md, csv and json:
```
adisos, line 149: why not also for `txt`?

test_outputs/connlist/exposure_acs-security-demos_connlist_output.csv
zivnevo: I don't see much value in reading exposure analysis results in CSV format.
Humans should better read the `txt` or the `md` version, while machines should read the `json` version.
So, any decision here is OK.

adisos: OK, and what about `md` ? (should split to ingress / egress sections, to be consistent with `txt`?)

pkg/netpol/connlist/conns_formatter_md.go
shireenf-ibm: i can instead of checking the exposure result's length , to check the exposure analysis flag (on with --exposure)
and if the result empty, to print header with empty section

adisos: ok, sounds good

test_outputs/connlist/exposure_acs-security-demos_connlist_output.csv
zivnevo: yes, better be consistent between `md` and `txt`

pkg/netpol/connlist/conns_formatter_md.go
shireenf-ibm: done

test_outputs/connlist/exposure_acs-security-demos_connlist_output.csv
shireenf-ibm: done, csv, json and md are consistent with txt now. 
i.e. exposure conns are separated into two sections : egress exposure and ingress exposure.

pkg/netpol/connlist/conns_formatter.go
shireenf-ibm: changes were made (with the consistency of all formats, these are common now also for txt)

adisos (approved):

PR #365, opened 2024-06-11
Handling selectors with match expression

shireenf-ibm:
#236
task:

> - [ ] support selectors with `matchExpression` (all operators)


shireenf-ibm:
since last merge was not successful will close this PR and open new one with only the relevant commits

PR #377, opened 2024-06-23
handling selectors with matchexpressions (fixed)

shireenf-ibm:
#236
task:

> - [ ] support selectors with `matchExpression` (all operators)

in this PR:
- added support of representative peers which are inferred from rules with `matchExpressions`
- it was determined that representative peers inferred from rules with `matchExpressions` will not be removed if there are real peer matching them (not redundant also: peers matching all-namespaces / peers matching all-pods in a namespace)
- it was also determined that when computing allowed connections, a rule matches a representative peer if :
   * the rule is empty (matches all)
   * the rule points to same reference of the selectors of the representative peer
   * the rule and the selectors of the representative peer have same requirements 
    (1 way containment is not a match)
- cancelled creating representative namespaces ( if the representative peer has the same namespace of a policy (inferred from a rule with nil `namespaceSelector` it will be generated in that namespace, otherwise it will have a nil namespace object
-  the representative namespace and pod selectors of a representative peer are stored in its Pod object
- the `RepresentativePeer` struct was eliminated , a representative peer is a `WorkloadPeer` with `kind == RepresentativePeer`
- changing the type of `NamespacLabels` and `PodLabels` of an `ExposedPeer` to be `LabelSelector` so it may include both `matchLabels` and `matchExpressions`

------------------

more things where done in this PR :
- fixing multiple `typos` in `netpol-analyzer`'s code
- renaming all new exposure tests dirs to start with `exposure_` , and renaming expected output's suffixes in this way :
  * if the test run with exposure-analysis flag off : then a connlist output is generated -> expected output suffix is `connlist_output.<format>`
  * if the test run with exposure-analysis flag on : the the suffix of expected output is `exposure_output.<format>` to identify that the output contains also exposure results.
  this way also helps to differentiate when a test runs with the flag on/off (even for `exposure_<test_dir>`  
 - renaming some functions/variables for more readability
 - updating some functions / in files documentations for clarity
 - unit tests for the new functionality
 - adding multiple new (connlist) tests with the new feature


pkg/netpol/connlist/conns_formatter.go
```diff
@@ -104,16 +114,28 @@ func formExposureItemAsSingleConnFiled(peerStr string, exposureItem XgressExposu
 }
 
 // convertLabelsMapToString returns a string representation of the given labels map
+// considers the special labels (with requirements such as Exists, DoesNotExist, NotIn)
 func convertLabelsMapToString(labelsMap map[string]string) string {
-	return labels.SelectorFromSet(labels.Set(labelsMap)).String()
+	labelsSrings := make([]string, 0)
+	for k, v := range labelsMap {
+		if v == common.ExistsVal {
```
adisos, line 121: can the `exists` keyword be imported from a pkg that handles selectors, instead of re-defining it in our `common` pkg?

adisos (commented):
few initial comments

adisos:
general comment: the expression selectors could be represented as `LabelSelectorRequirement` from pkg `go\pkg\mod\k8s.io\apimachinery@v0.29.2\pkg\apis\meta\v1\types.go`, instead of converting to `map[string]string` and then re-converting to a string.

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -104,16 +114,28 @@ func formExposureItemAsSingleConnFiled(peerStr string, exposureItem XgressExposu
...
+		}
+		if v == common.DoesNotExistVal {
+			labelsSrings = append(labelsSrings, k+space+doesNotExist)
+			continue
+		}
+		if strings.HasPrefix(v, common.NotPrefix) {
+			labelsSrings = append(labelsSrings, k+space+notIn+space+v[1:])
+			continue
+		}
+		labelsSrings = append(labelsSrings, k+equal+v)
```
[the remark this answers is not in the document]
shireenf-ibm: operator `in` was converted to a "matchlabels" version of <key:val> for each value in its values list.
and hence seperated to multiple representative peers ; so representative labels will definitely not include `in` operator

shireenf-ibm:
> general comment: the expression selectors could be represented as `LabelSelectorRequirement` from pkg `go\pkg\mod\k8s.io\apimachinery@v0.29.2\pkg\apis\meta\v1\types.go`, instead of converting to `map[string]string` and then re-converting to a string? why do we need our own conversion? could just add a separate field for such selectors of this type instead?


I thought about this either but found that it will cost multiple changes and will not differ that much for the "special" cases
1* we can add fields of "LabelSelectorRequirement" to the k8s.Pod and eval.RepresentativePeer interfaces and consider them each time needed; but still the `ExposedPeer` will contain the representative labels as `map[string]string (so preferred to be consistent from beginning) 
2* operator `In` case; I preferred to convert the expression into labels of <key:val> for each value (and so a new representative peer is created for each)
In case of the other operators it was only to convert the to a single <key>:<val> where the <val> is "special" and needs to be "compared" specifically , (same comparisons  would be done in case we use the `LabelSelectorRequirement` )

shireenf-ibm:
some comments regarding the last [commit](https://github.com/np-guard/netpol-analyzer/pull/377/commits/31e1ec4595f9cee6046c46f4e91162bde245f661):
current code changes:
- since the potential namespaces and pods a peer may be exposed  to can contain either labels map or list of requirements 
I used the [LabelSelector](https://pkg.go.dev/k8s.io/apimachinery@v0.29.2/pkg/apis/meta/v1#LabelSelector) struct to represent them.(it includes both `map[string]string` and `[]LabelSelectorRequirement`)

- however , in order to print the potential labels and requirements string I prefered to use the labels string func and LabelSelectorRequirement.String() func rather than the LabelSelector.String() ; in order to get a neater output.

- even though `LabelSelector` struct includes both labels and requirements fields; in `k8s.pod` and `k8s.namespace` kept two different fields for each (since real pods and namespaces has only labelsMap)

#### a suggestion: (not implemented here)

- since `k8s.namespace` contains both requirements and labels; we have the option to remove the `RepresentativePeer struct` and use `PodPeer struct` to represent representative peer, by differentiating the `kind` value of the peer; (if this is desired, i prefer to do in a different PR)

test_outputs/connlist/exposure_test_egress_exposure_with_named_port_connlist_output.dot.png
adisos, line 1: can you also duplicate these tests for scenarios with actual pods also matched by these selectors?

pkg/netpol/connlist/conns_formatter.go
shireenf-ibm: not relevant anymore

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -104,16 +114,28 @@ func formExposureItemAsSingleConnFiled(peerStr string, exposureItem XgressExposu
 }
 
 // convertLabelsMapToString returns a string representation of the given labels map
+// considers the special labels (with requirements such as Exists, DoesNotExist, NotIn)
 func convertLabelsMapToString(labelsMap map[string]string) string {
-	return labels.SelectorFromSet(labels.Set(labelsMap)).String()
+	labelsSrings := make([]string, 0)
+	for k, v := range labelsMap {
+		if v == common.ExistsVal {
+			labelsSrings = append(labelsSrings, k+space+exists)
```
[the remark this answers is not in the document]
shireenf-ibm: not relevant anymore

pkg/netpol/connlist/conns_formatter.go
```diff
@@ -77,11 +75,23 @@ func formSingleP2PConn(conn Peer2PeerConnection) singleConnFields {
...
+	mapOpen                = "{"
+	mapClose               = "}"
+	equal                  = "="
+	comma                  = ","
+	key                    = "key"
+	colon                  = ": "
+	space                  = " "
+	notIn                  = "NotIn"
+	doesNotExist           = "DoesNotExist"
+	exists                 = "Exists"
```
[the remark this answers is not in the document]
shireenf-ibm: not relevant anymore

test_outputs/connlist/exposure_test_egress_exposure_with_named_port_connlist_output.dot.png
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -208,7 +208,18 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
 					return false, err
 				}
 				peerNamespace := peer.GetPeerNamespace()
-				peerMatchesNamespaceSelector = selector.Matches(labels.Set(peerNamespace.Labels))
+				// checking if a selector matches labels by peer type; since representative peers selectors may need special handling
+				if isRepresentativePod(peer) {
+					// representative peer's namespace labels may be inferred from a rule with special matchExpression requirements
+					// and also contains the representative ns name label which is not relevant for comparison
+					peerMatchesNamespaceSelector, err = SelectorMatchesRepresentativePeerLabels(selector, peerNamespace.Labels,
```
adisos, line 215: can't we just keep a reference to the selector(s) from which this representative peer was created,  and consider a match only if this is the relevant selecotr, instead of implementing containment of selectors comparison?

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,175 @@
...
+// - if the requirement operator is NotIn :
+// * the operator of the same key should not be Exists (notIn and exists are not equiv; notIn selects also doesNotExist)
+// * if the operator of the same key in the representative labels is NotIn; then the list of values of the requirement needs
+// to be contained in the !(<values>) list
+// or the representative labels should Match the requirement (i.e. key does not exist/ key with value not from the list)
+// ******************************************************************************************************* //
+
+// SelectorMatchesRepresentativePeerLabels given a selector from a policy rule and labels and requirements of representative peer;
+// check if the selector matches the repreaentative selector created by the peer's labels and requirements
+func SelectorMatchesRepresentativePeerLabels(selector labels.Selector, pLabels map[string]string,
```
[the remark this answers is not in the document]
shireenf-ibm: 1. right, it does not have to be exported (fixing)
2. will add
3. in the older implementation of this, I did use an instance of labels to represent the requirements (comparison was not simpler/ much different).
 the imported packages of k8s, don't "convert" requirements of matchExpressions into labels (unless it is IN operator with only one value) and thus in both cases I need to implement a comparing subset (considering the values/ operators)

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: I think we still need this containment, some selectors may intersect / be equivalent on one way to other selectors 
i.e. a representative peer (built from one selector) may match two or more selectors actually and then the connection contains the ports of all the selectors. 
an example : `test_exposure_with_different_rules_6` (there are more examples like this too)

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,72 @@
...
+		return false, err
+	}
+	if selector1.Empty() { // empty rule matches everything
+		return true, nil
+	}
+	selector2, err := v1.LabelSelectorAsSelector(repSelector)
+	if err != nil {
+		return false, err
+	}
+	requirements1, _ := selector1.Requirements() // Requirements() returns sorted by key list
```
adisos, line 40: where is it documented that `Requirements()` returns a sorted list?

pkg/netpol/eval/internal/k8s/namespace.go
```diff
@@ -16,6 +17,10 @@ import (
 type Namespace struct {
 	Name   string
 	Labels map[string]string
+	// RepresentativeNsLabelSelector points to the namespaceSelector of the policy rule which this
+	// representative namespace was inferred from
+	// used only with representative peers (exposure-analysis)
+	RepresentativeNsLabelSelector *v1.LabelSelector
```
adisos, line 23: why is this part of the namespace object, and not as part of the representative pod object?

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: I found this by reading the implementation.
`LabelSelectorAsSelector` builds a new `labels.selector` by using `Requirements.Add` which sorts the requirements slice by key. 
and `Requirements()` returns this list 
links:
[Add](https://github.com/kubernetes/apimachinery/blob/d7e1c5311169d5ece2db0ae0118066859aa6f7d8/pkg/labels/selector.go#L373)
[LabelSelectorAsSelector](https://github.com/kubernetes/apimachinery/blob/dc7e034c86479d49be4b0eefad307621e10caa0e/pkg/apis/meta/v1/helpers.go#L34)

adisos: - please add links in the comment about the expected sorted result. 
- please add unit tests to the functions in this file.

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -519,54 +536,39 @@ func (np *NetworkPolicy) scanEgressRules() ([]SingleRuleLabels, error) {
...
+		ruleSel.PodSelector = rules[i].PodSelector
+		ruleSel.NsSelector = rules[i].NamespaceSelector
+		if rules[i].NamespaceSelector == nil {
+			// special case: ns selector is nil but podSelector is not, so the namespace of the rule is
+			// the policy's namespace; (turn on the indicator)
 			ruleSel.PolicyNsFlag = true
-			ruleSel.NsLabels = map[string]string{common.K8sNsNameLabelKey: np.Namespace}
-		} else {
-			ruleSel.NsLabels = nsSelectorMap
+			nsNameLabelSel := map[string]string{common.K8sNsNameLabelKey: np.Namespace}
```
adisos, line 564: can avoid duplicating the map of single label containing `K8sNsNameLabelKey` ? there is another function building this map currently

pkg/netpol/eval/exposure.go
```diff
@@ -30,18 +31,20 @@ func generateNewPodName(index int) string {
...
-	for i := range selectorsLabels {
+// - generated representative peers are unique; i.e. if different rules (e.g in different policies or different directions)
+// has same labels, one representative peer is generated to represent both
+func (pe *PolicyEngine) generateRepresentativePeers(selectors []k8s.SingleRuleSelectors, policyName, policyNs string) (err error) {
+	for i := range selectors {
 		// if ns labels of the rule selector was nil, then the namespace of the pod is same as the policy's namespace
-		if selectorsLabels[i].PolicyNsFlag {
-			_, err = pe.AddPodByNameAndNamespace(generateNewPodName(i), policyNs, &selectorsLabels[i])
+		if selectors[i].PolicyNsFlag {
+			_, err = pe.AddPodByNameAndNamespace(generateNewPodName(i), policyNs, &selectors[i])
```
adisos, line 44: consider renaming `AddPodByNameAndNamespace` ? split to the various cases as different functions?

pkg/netpol/eval/exposure.go
```diff
@@ -52,29 +55,38 @@ func (pe *PolicyEngine) generateRepresentativePeers(selectorsLabels []k8s.Single
 
 // extractLabelsAndRefineRepresentativePeers extracts the labels of the given pod object and its namespace and refine matching peers
 // helping func - added in order to avoid code dup. in upsertWorkload and upsertPod
-func (pe *PolicyEngine) extractLabelsAndRefineRepresentativePeers(podObj *k8s.Pod) error {
+func (pe *PolicyEngine) extractLabelsAndRefineRepresentativePeers(podObj *k8s.Pod) {
```
adisos, line 58: please explain in documentation what it means to "refine" -- delete a representative pod if there is a real pod that matches its selectors (matchLabels only, no matchExpression).

shireenf-ibm: done

shireenf-ibm:
attaching a list of the tests that were added in this PR, later will update the file with all tests of exposure-analysis
[exposure_analysis_tests.csv](https://github.com/user-attachments/files/16233553/exposure_analysis_tests.csv)

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: > * please add links in the comment about the expected sorted result.
 done

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,72 @@
...
+
+// this file handles checking equality of representative peers' selectors' requirements
+// since a representative peer is a peer inferred from a network-policy rule
+
+// areRequirementsEqual given two labelSelector objects, one from a policy rule and the other from a represenatative pod/namespace,
+// returns true if:
+// 1. both selectors point to same reference (rule and its matching representative pod/ns)
+// 2. if the rule's selector is empty (matches all pods/namespaces)
+// 3. if the requirements of both rules are equal (same)
+func areRequirementsEqual(ruleSelector, repSelector *v1.LabelSelector) (bool, error) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/exposure.go
shireenf-ibm: done.
1. reverted all changes to `AddPodByNameAndNamespace ` (to the main branch version) so it applies only for ingress-controller cases (renaming this func, if neede - should be done in the ingress-analysis branch, not here)
2. added a new function  `addRepresentativePod` for our use-case (exposure-analysis).

pkg/netpol/eval/internal/k8s/namespace.go
shireenf-ibm: moved to the pod object

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: >     * please add unit tests to the functions in this file.

done

adisos (commented):
did not review all changes yet, added few questions

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -60,6 +60,12 @@ type Pod struct {
 	// - and the maximal connection-set for which the pod is exposed to all namespaces by network policies
 	// on egress direction
 	EgressExposureData PodExposureInfo
+	// RepresentativePodLabelSelector contains reference to the podSelector of the policy-rule which the representative peer was inferred from
+	// used only with representative Pods
+	RepresentativePodLabelSelector *v1.LabelSelector
```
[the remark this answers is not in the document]
shireenf-ibm: > are these fields expected to be nil if and only if this is not a representative pod? if the selectors are not nil, are they expected to be not empty?

no. 
a `RepresentativePodLabelSelector` may be nil also for representative pod (a nil podSelector in a policy rule means all pods)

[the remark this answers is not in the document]
shireenf-ibm: > shouldn't these fields capture a set of references? (could have same selectors in several policies?)

 we may have same selectors in same policy in the different directions, and in different policies
 same selectors also may have same requirements in different orders
 
 I don't find that storing set of references will defer a lot of storing the *LabelSelector  > both will work
and  same functions still needed to be used (to compare the requirements)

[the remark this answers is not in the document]
shireenf-ibm: actually it me be more efficient, by using the hash key , I can avoid comparing requirements later 
will change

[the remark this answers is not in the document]
adisos: > a RepresentativePodLabelSelector may be nil also for representative pod
can you add documentation about that ?
also, if RepresentativePodLabelSelector  is nil, then at least ns selector expected to be not nil? 
and what about empty selectors?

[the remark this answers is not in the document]
shireenf-ibm: I will update the documentations with this: ( and will update relevant the places in code.)
```
 possible combinations of RepresentativePodLabelSelector and RepresentativeNsLabelSelector:
	 - both are specific non-empty selectors
	 - RepresentativePodLabelSelector is nil + RepresentativeNsLabelSelector is a specific non-empty selector
	 - RepresentativePodLabelSelector is empty + RepresentativeNsLabelSelector is a specific non-empty selector
	 - RepresentativePodLabelSelector a specific non-empty selector + RepresentativeNsLabelSelector is an empty selector

 both might not be nil / empty at same time
```

 
- if a rule contains a nil `namespaceSelector `:  `RepresentativeNsLabelSelector` will contain a generated selector with matchLabels with the namespace key and the namespace of the policy 
- if a rule contains an empty/nil `podSelector` + empty `namespaceSelector `: this represents an entire-cluster connection that will be represented in the general conns of the netpol itself (and will not be presented by any representative peer)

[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -80,16 +86,18 @@ func PodFromCoreObject(p *corev1.Pod) (*Pod, error) {
...
+		Labels:                         make(map[string]string, len(p.Labels)),
+		IPs:                            make([]corev1.PodIP, len(p.Status.PodIPs)),
+		Ports:                          make([]corev1.ContainerPort, 0, defaultPortsListSize),
+		HostIP:                         p.Status.HostIP,
+		Owner:                          Owner{},
+		FakePod:                        false,
+		IngressExposureData:            initiatePodExposure(),
+		EgressExposureData:             initiatePodExposure(),
+		RepresentativePodLabelSelector: nil,
+		RepresentativeNsLabelSelector:  nil,
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -214,13 +222,15 @@ func PodsFromWorkloadObject(workload interface{}, kind string) ([]*Pod, error) {
 		pod.FakePod = false
 		pod.IngressExposureData = initiatePodExposure()
 		pod.EgressExposureData = initiatePodExposure()
+		pod.RepresentativePodLabelSelector = nil
+		pod.RepresentativeNsLabelSelector = nil
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/resources.go
```diff
@@ -584,38 +578,61 @@ func (pe *PolicyEngine) ConvertPeerNamedPort(namedPort string, peer Peer) (int32
...
 }
+
+// addRepresentativePod adds a new representative pod to the policy-engine (to pe.representativePeersMap)
+// if the given namespace string (ns) is not empty (i.e. a real namespace name), the representative pod is created in that namespace
+// else, the representative pod will have no namespace (will not add a representative namespace to the policy-engine)
+// and its "namespace" requirements will be stored in its RepresentativeNsLabelSelector field to be used when needed
+// used only in exposure-analysis
+func (pe *PolicyEngine) addRepresentativePod(name, ns string, objSelectors *k8s.SingleRuleSelectors) error {
+	var nsSelector, podSelector *metav1.LabelSelector
+	if objSelectors != nil {
```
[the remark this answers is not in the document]
shireenf-ibm: true this was relevant only when used same `func` for both cases (ingress-controller and representative pod)
**fixed** (now returning an error if the `objSelectors ` is nil; even we shouldn't get here at all)

pkg/netpol/eval/resources.go
```diff
@@ -584,38 +578,61 @@ func (pe *PolicyEngine) ConvertPeerNamedPort(namedPort string, peer Peer) (int32
...
+			return err
+		}
+	}
+
+	//  compute a unique string from the label selectors to be used as the map key
+	nsKey, err := k8s.VariantFromLabelsSelector(nsSelector)
+	if err != nil {
+		return err
+	}
+	podKey, err := k8s.VariantFromLabelsSelector(podSelector)
```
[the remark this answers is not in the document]
shireenf-ibm: a `objSelectors` might not be nil
nsSelector is always not nil so at least the `nsKey` is always specific (not 0)

pkg/netpol/eval/check.go
```diff
@@ -79,10 +79,16 @@ func (pe *PolicyEngine) convertPeerToPodPeer(peer Peer) (*k8s.PodPeer, error) {
 	return podPeer, nil
 }
 
-func (pe *PolicyEngine) getPeerNamespaceObject(namespaceName, peerStr string) (*k8s.Namespace, error) {
+func (pe *PolicyEngine) getPeerNamespaceObject(namespaceName, peerName string) (*k8s.Namespace, error) {
+	// if this is a representative peer which is not in a real namespace return nil;
+	// PolicyEngine does not contain representative namespaces.
+	if namespaceName == "" && strings.HasPrefix(peerName, k8s.RepresentativePodName) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

shireenf-ibm:
some review tasks:

1. **eval pkg:**

- [ ] RepresentativePeer - type : consider eliminate, and use WorkloadPeer with kind() instead 

- [ ] func (pe *PolicyEngine) AddObjects :

-  consider renaming, "AddObjectsByOrder", "AddObjectsForExposureAnalysis"

-  documentation 

- [ ] fix typos in documentation/code (for example pkg\netpol\eval\resources.go ) 

- [ ]  func (pe *PolicyEngine) AddObjects(objects []parser.K8sObjec : 
why should we add namespaces in the beginning, and not only policy objects? 

- [ ]  if needed to add namespaces, replace comment:
 "// 2. namespaces: so when upserting workloads, we'll be able to refine correctly representativePeers with
// namespace name/ labels similar to those belonging the workloads' namespace "
 => " check if a generated representative peer should be removed, if its labels and namespace correspond to a real pod " 

- [ ] in function :

- func (pe *PolicyEngine) addObjectsByKind(objects []parser.K8sObject) error ,
   why:
	```
          if !pe.exposureAnalysisFlag { // for exposure analysis; this already done
		return pe.resolveMissingNamespaces()
	}
       ```

- func addRepresentativePod : remove call to resolveSingleMissingNamespace() ??

- resolveSingleMissingNamespace() : change so that it is called once instead of twice in different places for exposure analysis on/off 

- [ ] ScanPolicyRulesForGeneralConnsAndRepresentativePeers -> consider renaming:

"ScanPolicyRulesAndUpdateGeneralConns" 

- [ ] "GeneralConns" -> consider renaming to "ExposedGeneralConns" / "ExposedWideConns" 

- [ ] "handleRulesSelectors" -> consider renaming to 

"getSelectorsAndUpdateGeneralConns(FromRules)"



- [ ] consider removing `PolicyNsFlag`



- [ ] generateRepresentativePeers() => generateNewPodName().. 

-  fix naming, no need for index

- add comment explaining why naming is not important / used 


- [ ] "extractLabelsAndRefineRepresentativePeers" -> 

rename to "RemoveRedundantRepresentativePeers" 
// when possible, remove... 

2. **connlist.go:**

- [ ] connPeers => consider renaming. ("allPeers", "realAndFakePeers", "realAndRepresentativePeers")

shireenf-ibm:
> 
>     * [ ]  in function :
> 
>     * func (pe *PolicyEngine) addObjectsByKind(objects []parser.K8sObject) error ,
>       why:
>       ```
>              if !pe.exposureAnalysisFlag { // for exposure analysis; this already done
>        	return pe.resolveMissingNamespaces()
>        }
>           ```
>       ```
> 
>     * resolveSingleMissingNamespace() : change so that it is called once instead of twice in different places for exposure analysis on/off
> 
>     * [ ]   func (pe *PolicyEngine) AddObjects(objects []parser.K8sObjec :
>       why should we add namespaces in the beginning, and not only policy objects?
> 

hi @adisos , 
regarding splitting namespaces with policies in `AddObjectsForExposureAnalysis` and having the `resolveSingleMissingNamespace` calls twice (when exposure-analysis is on/off).

I found that we really should split namespaces at the beginning (with policies)

I have committed an example that will give us a misleading result if the namespaces were not split at the beginning.

in the example we have:
1. a real namespace in the manifests with label `name:ns2` and we have a real pod in that namespace.
2. the netpol's rule exactly match the namespace and pod labels (of the namespace and the pod).

here are the results with and without splitting the namespaces :
1. with splitting the namespaces, we get the correct result (no exposure inferred from the rule since we have an exact match):
results_with_namespaces_split

2. without split the namespaces, we get an exposure result with the unnecessary line:
`hello-world/workload-a[Deployment]      <=      [namespace with {name=ns2}]/[pod with {app=b-app}] : All Connections`
results_without_namespaces_split



/////////// an explanation:
> 
>     * [ ]   func (pe *PolicyEngine) AddObjects(objects []parser.K8sObjec :
>       why should we add namespaces in the beginning, and not only policy objects?
> 
- if we don't split the namespaces at the beginning, but the policy engine gets the `pod` object first (the parsed yaml object); 
then the policy engine will resolve the missing namespace of the `pod` with only the label of default k8s name (which is not `name`).
and real namespace labels of the real pod will not contain the actual real labels that are in the `namespace object` yaml.
and the redundant matching representative peer won't be removed.

- however if we split real `namespaces` objects from the manifests at the beginning (with the policies);
then the namespace with all its real labels will be in the policy-engine when we get to the real pod and those labels will be considered and the redundant representative-peer will be removed.

///////////////////////
regarding calling the `resolveSingleMissingNamespace` twice:
> 
>     * [ ]  in function :
> 
>     * func (pe *PolicyEngine) addObjectsByKind(objects []parser.K8sObject) error ,
>       why:
>       ```
>              if !pe.exposureAnalysisFlag { // for exposure analysis; this already done
>        	return pe.resolveMissingNamespaces()
>        }
>           ```
>       ```
> 
>     * resolveSingleMissingNamespace() : change so that it is called once instead of twice in different places for exposure analysis on/off
> 

-- so when the exposure-analysis is on : the namespaces are inserted to the policy-engine first; and so we can use the func 
`resolveSingleMissingNamespace` when inserting a pod/workload; because if its namespace not found in the policy-engine; it is for sure not found in the resources too 

-- but when exposure-analysis is off : when adding a pod/workload to policy-engine; its namespace might be in the resources but was not added yet (the objects list is not sorted and we can not predict what resource comes first); 
so we can't call `resolveSingleMissingNamespace` until all k8s objects from the resource were added to the policy-engine 
(otherwise, a real namespace with some specified labels for it may be missed)

pkg/netpol/eval/exposure.go
```diff
@@ -18,63 +18,71 @@ import (
...
-		if selectorsLabels[i].PolicyNsFlag {
-			_, err = pe.AddPodByNameAndNamespace(generateNewPodName(i), policyNs, &selectorsLabels[i])
-		} else {
-			_, err = pe.AddPodByNameAndNamespace(generateNewPodName(i), generateNewNamespaceName(policyName, i), &selectorsLabels[i])
+// has namespace labelSelector and pod labelSelector inferred from single entry of selectors in a policy rules list;
+//
+// - for example, if a rule within policy has an entry: namespaceSelector "foo: managed", then a representative pod
+// with this labelSelector will be added, representing all potential pods in such a namespace.
+// - generated representative peers are unique; i.e. if different rules (e.g in different policies or different directions)
+// has same selectors, one representative peer is generated to represent both.
```
adisos, line 27:
```suggestion
// have same selectors, one representative peer is generated to represent both.
```

pkg/netpol/eval/exposure.go
```diff
@@ -18,63 +18,71 @@ import (
...
-// helping func - added in order to avoid code dup. in upsertWorkload and upsertPod
-func (pe *PolicyEngine) extractLabelsAndRefineRepresentativePeers(podObj *k8s.Pod) error {
-	// since namespaces are already upserted; if pod's ns not existing resolve it
-	if _, ok := pe.namspacesMap[podObj.Namespace]; !ok {
-		// the "kubernetes.io/metadata.name" is added automatically to the ns; so representative peers with such selector will be refined
-		err := pe.resolveSingleMissingNamespace(podObj.Namespace, nil)
-		if err != nil {
-			return err
-		}
+// removeRedundantRepresentativePeers extracts the labels of the given pod object and its namespace and refine matching
```
adisos, line 48:
```suggestion
// removeRedundantRepresentativePeers removes redundant representative peers, given an input real pod.
// The motivation is that exposure analysis will not report exposure to a representative-pod that has selectors
// exactly matched by a real pod. (This removal applies only to selectors of matchLabels, not matchExpression).
```

pkg/netpol/eval/exposure.go
```diff
@@ -18,63 +18,71 @@ import (
...
+// representative-peers, i.e. delete a representative pod if the given real pod matches its selectors
```
adisos, line 49:
```suggestion
```

pkg/netpol/eval/exposure.go
```diff
@@ -18,63 +18,71 @@ import (
...
+// (applied for representative-peers with matchLabels only, no matchExpression).
```
adisos, line 50:
```suggestion
```

pkg/netpol/eval/exposure.go
```diff
@@ -18,63 +18,71 @@ import (
...
+// helping func - added in order to avoid code dup. in insertWorkload and insertPod
```
adisos, line 51:
```suggestion
```

pkg/netpol/eval/exposure.go
```diff
@@ -131,25 +140,37 @@ func (pe *PolicyEngine) GetPeerXgressEntireClusterConn(p Peer, isIngress bool) (
...
+	peer, err := isPeerAWorkloadPeer(p)
+	if err != nil {
+		return v1.LabelSelector{}, v1.LabelSelector{}, err
+	}
+	if peer.Kind() != k8s.RepresentativePeerKind { // should not get here
+		return v1.LabelSelector{}, v1.LabelSelector{}, errors.New(netpolerrors.NotRepresentativePeerErrStr(p.String()))
+	}
+	podLabels = v1.LabelSelector{}
+	if peer.Pod.RepresentativePodLabelSelector != nil {
+		podLabels = *peer.Pod.RepresentativePodLabelSelector.DeepCopy()
```
adisos, line 169: why DeepCopy is required here?

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -29,30 +29,32 @@ import (
...
 // @todo might help if while pre-process, to check containment of all rules' connections; if all "specific" rules
 // connections are contained in the "general" rules connections, then we can avoid iterating policy rules for computing
 // connections between two peers
 
-type PolicyGeneralRulesConns struct {
-	// AllDestinationsConns contains the maximal connection-set which the policy's rules allow to all destinations
+// PolicyExposureWithoutSelectors describes the maximum allowed conns that the policy exposes as cluster wide or as external,
+// those conns are inferred when the policy has no rules, or from empty rules
+type PolicyExposureWithoutSelectors struct {
+	// ExternalExposure  contains the maximal connection-set which the policy's rules allow from/to external destinations
```
adisos, line 53:
```suggestion
	// ExternalExposure  contains the maximal connection-set which the policy's rules allow from/to external and cluster-wide destinations
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -512,79 +533,61 @@ func (np *NetworkPolicy) scanEgressRules() ([]SingleRuleLabels, error) {
...
 }
 
-// isEmptyMap gets a labels map and returns if it is not nil but empty map
-func isEmptyMap(labelsMap map[string]string) bool {
-	return labelsMap != nil && len(labelsMap) == 0
-}
-
-func (np *NetworkPolicy) updateNetworkPolicyGeneralConn(allDests, entireCluster bool, rulePorts []netv1.NetworkPolicyPort,
+// updateNetworkPolicyWideExposureConns updates the wide exposure connections of the policy
+func (np *NetworkPolicy) updateNetworkPolicyWideExposureConns(externalExposure, entireCluster bool, rulePorts []netv1.NetworkPolicyPort,
```
adisos, line 573: rename func to `updateNetworkPolicyExposureClusterWideConns`

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -512,79 +533,61 @@ func (np *NetworkPolicy) scanEgressRules() ([]SingleRuleLabels, error) {
...
 	return rulesSelectors, nil
 }
 
-// isEmptyMap gets a labels map and returns if it is not nil but empty map
-func isEmptyMap(labelsMap map[string]string) bool {
-	return labelsMap != nil && len(labelsMap) == 0
-}
-
-func (np *NetworkPolicy) updateNetworkPolicyGeneralConn(allDests, entireCluster bool, rulePorts []netv1.NetworkPolicyPort,
+// updateNetworkPolicyWideExposureConns updates the wide exposure connections of the policy
```
adisos, line 572:
```suggestion
// updateNetworkPolicyWideExposureConns updates the cluster-wide exposure connections of the policy
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -126,9 +128,9 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
 			if err != nil {
 				return res, err
 			}
-			if (dst == nil || isRepresentativePod(dst)) && portName != "" {
-				// adding namedPort to connectionSet in case of :
-				// - dst is nil - for general connections;
+			if (dst == nil || hasRepresentativePod(dst)) && portName != "" {
+				// adding namedPort string to connectionSet in case of :
+				// - dst is nil - for updating wide exposure connections;
```
adisos, line 133:
```suggestion
				// - dst is nil - for updating cluster-wide exposure connections;
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -126,9 +128,9 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
 			if err != nil {
 				return res, err
 			}
-			if (dst == nil || isRepresentativePod(dst)) && portName != "" {
-				// adding namedPort to connectionSet in case of :
-				// - dst is nil - for general connections;
+			if (dst == nil || hasRepresentativePod(dst)) && portName != "" {
+				// adding namedPort string to connectionSet in case of :
```
adisos, line 132: shouldn't check that exposure analysis is active?
can add more details here? why it is not added in other cases?

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -200,27 +202,42 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
...
+			var err error
+			var selector labels.Selector
 			if rulePeers[i].NamespaceSelector == nil {
 				peerMatchesNamespaceSelector = (np.ObjectMeta.Namespace == peer.GetPeerPod().Namespace)
 			} else {
-				selector, err := np.parseNetpolLabelSelector(rulePeers[i].NamespaceSelector)
+				peerNamespace := peer.GetPeerNamespace()
+				// checking if a selector matches labels by peer type; since representative peers selectors may need special handling
+				if hasRepresentativePod(peer) {
+					// representative peer is inferred from a rule with its/same labelSelector.
```
adisos, line 213: what do you mean by "its/same" ?

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -451,21 +473,18 @@ func (np *NetworkPolicy) fullName() string {
...
+	PodSelector *metav1.LabelSelector
 }
 
-// ScanPolicyRulesForGeneralConnsAndRepresentativePeers scans policy rules and :
-// - updates policy's general connections with all destinations or/ and with entire cluster for ingress and/ or egress directions
-// - returns list of selectors labels from rules which has non-empty namespace selectors  (from rules with namespaceSelector only)
-// TODO: support checking rules with podSelector
-func (np *NetworkPolicy) ScanPolicyRulesForGeneralConnsAndRepresentativePeers() (rulesSelectors []SingleRuleLabels, err error) {
+// ScanPolicyRulesAndUpdateExposedWideConns scans policy rules and :
+// - updates policy's exposed wide connections from/to external resources or/and from/to all namespaces in the cluster on
```
adisos, line 484:
```suggestion
// - updates policy's exposed cluster-wide connections from/to external resources or/and from/to all namespaces in the cluster on
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -451,21 +473,18 @@ func (np *NetworkPolicy) fullName() string {
...
+// ingress and egress directions
+// - returns list of labels.selectors from rules which has non-empty selectors, for which the representative peers should be generated
```
adisos, line 486:
```suggestion
// - returns list of labels.selectors from rules which have non-empty selectors, for which the representative peers should be generated
```

pkg/netpol/eval/internal/k8s/peer.go
```diff
@@ -51,23 +51,14 @@ type WorkloadPeer struct {
 	Pod *Pod
 }
 
-// RepresentativePeer implements eval.Peer interface
-// a representative peer is a peer inferred from a policy rule (selector) not a parsed pod/deployment object
+const podKind = "Pod"
+
+// A WorkloadPeer with kind == RepresentativePeer is a representativePeer
+// a representative peer is a peer inferred from a policy rule (selectors) not a parsed pod/deployment object
```
adisos, line 57:
```suggestion
// a representative peer is a peer inferred from a policy rule (selectors), not a parsed pod/deployment object
```

pkg/netpol/eval/internal/k8s/peer.go
```diff
@@ -51,23 +51,14 @@ type WorkloadPeer struct {
...
 // and is used to represent a potential pod/ns entity in the cluster (that does not exist on the input resources)
 // but may have enabled connectivity to input resources (pods/deployments) based on input network policies.
```
adisos, line 59:
```suggestion
// but may have permitted connectivity to input resources (pods/deployments) based on input network policies.
```

pkg/netpol/eval/internal/k8s/peer.go
```diff
@@ -91,7 +85,7 @@ func (p *WorkloadPeer) Kind() string {
 }
 
 func (p *WorkloadPeer) String() string {
```
adisos, line 87: is this function valid if `p` is representative peer?

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -52,14 +51,30 @@ type Pod struct {
 
 	// IngressExposureData contains:
```
adisos, line 52:
```suggestion
	// The fields below are relevant to real pods when exposure analysis is active:
    
    // IngressExposureData contains:
```

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -52,14 +51,30 @@ type Pod struct {
...
 	// EgressExposureData contains:
 	// - whether the pod is protected by any network-policy on egress direction or not;
-	// - and the maximal connection-set for which the pod is exposed to all namespaces by network policies
-	// on egress direction
+	// - the maximal connection-set for which the pod is exposed to all namespaces in the cluster by
+	// network policies on egress direction
 	EgressExposureData PodExposureInfo
+	// RepresentativePodLabelSelector contains reference to the podSelector of the policy-rule which the representative peer was inferred from
+	// used only with representative Pods
+	// RepresentativePodLabelSelector might be nil/ empty selector / a specific non-empty selector
```
adisos, line 64:
```suggestion
	// RepresentativePodLabelSelector might be nil/ empty selector / a specific non-empty selector

    // The fields below are relevant only to representative pod: 
```

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -52,14 +51,30 @@ type Pod struct {
...
+	RepresentativePodLabelSelector *v1.LabelSelector
+	// RepresentativeNsLabelSelector points to the namespaceSelector of the policy rule which this representative pod was inferred from
+	// used only with representative peers (exposure-analysis)
+	// RepresentativeNsLabelSelector might represent an empty selector / a specific non-empty selector (might not be nil)
```
adisos, line 68: "might not be nil" or "will not be nil" ?

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -52,14 +51,30 @@ type Pod struct {
...
+	// nil namespaceSelector in a policy-rule will be converted to the namespace name label when creating the representative pod.
+	RepresentativeNsLabelSelector *v1.LabelSelector
+
+	// possible combinations of RepresentativePodLabelSelector and RepresentativeNsLabelSelector:
+	// - both are specific non-empty selectors
+	// - RepresentativePodLabelSelector is nil + RepresentativeNsLabelSelector is a specific non-empty selector
```
adisos, line 74: please explain the meaning of each combintation.. for example: PodLabelSelector is nil + NsLabelSelector is a specific non-empty selector - does this imply all pods in the namespaces that matche this selector?

pkg/netpol/eval/resources.go
```diff
@@ -31,22 +31,22 @@ type (
...
 		podsMap                         map[string]*k8s.Pod                      // map from pod name to pod object
 		netpolsMap                      map[string]map[string]*k8s.NetworkPolicy // map from namespace to map from netpol name to its object
 		podOwnersToRepresentativePodMap map[string]map[string]*k8s.Pod           // map from namespace to map from pods' ownerReference name
 		// to its representative pod object
 		cache                  *evalCache
 		exposureAnalysisFlag   bool
-		representativePeersMap map[string]*k8s.RepresentativePeer // map from unique labels string to representative peer object,
-		// used only with exposure analysis
+		representativePeersMap map[string]*k8s.WorkloadPeer // map from unique labels string to representative peer object,
+		// used only with exposure analysis // representative peer object is a workloadPeer with kind == "RepresentativePeer"
```
adisos, line 42:
```suggestion
		// used only with exposure analysis (representative peer object is a workloadPeer with kind == "RepresentativePeer")
```

pkg/netpol/eval/resources.go
```diff
@@ -157,30 +163,35 @@ func (pe *PolicyEngine) addObjectsByKind(objects []parser.K8sObject) error {
...
 
-// resolveSingleMissingNamespace create a ns object and upsert to PolicyEngine
-func (pe *PolicyEngine) resolveSingleMissingNamespace(ns string, nsLabels map[string]string) error {
-	nLabels := nsLabels
-	if len(nLabels) == 0 {
-		nLabels = map[string]string{
-			common.K8sNsNameLabelKey: ns,
-		}
+// addMissingNamespace - helping func (to avoid duplicates); checks if the ns is not in the namespacesMap, resolves it
+func (pe *PolicyEngine) addIfMissingNamespace(ns string) (err error) {
```
adisos, line 174: this func can be part of the `resolveSingleMissingNamespace` func?

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -451,21 +473,18 @@ func (np *NetworkPolicy) fullName() string {
...
+func (np *NetworkPolicy) ScanPolicyRulesAndUpdateExposedWideConns() (rulesSelectors []SingleRuleSelectors, err error) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -512,79 +533,61 @@ func (np *NetworkPolicy) scanEgressRules() ([]SingleRuleLabels, error) {
...
+// getSelectorsAndUpdateExposedGeneralConns:
+// loops given rules list:
+// - if the rules list is empty updates the external and cluster wide exposure of the policy
+// - if a rule have empty selectors (podSelector and namespaceSelector)or just empty namespaceSelector (nil podSelector)
+// updates the cluster wide exposure of the policy
+// - if a rule contains at least one defined selector : appends the rule selectors to a selector list which will be returned.
 // this func assumes rules are legal (rules correctness check occurs later)
-func (np *NetworkPolicy) doRulesExposeToAllDestOrEntireCluster(rules []netv1.NetworkPolicyPeer, rulePorts []netv1.NetworkPolicyPort,
-	isIngress bool) (rulesSelectors []SingleRuleLabels, err error) {
+func (np *NetworkPolicy) getSelectorsAndUpdateExposedGeneralConns(rules []netv1.NetworkPolicyPeer, rulePorts []netv1.NetworkPolicyPort,
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/resources.go
adisos, line 1: could not understand the function `updatePodOwnersToRepresentativePodMapIfRequired` in this file

pkg/netpol/eval/resources.go
```diff
@@ -574,48 +585,76 @@ func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace st
...
+	}
+	nsLabelSelector := objSelectors.NsSelector
+	if nsLabelSelector == nil { // equal to podNs != ""
+		// if the objSelectors.NsSelector is nil, means inferred from a rule with nil nsSelector, which means the namespace of the
+		// pod is the namespace of the policy, so adding it as its RepresentativeNsLabelSelector requirement.
+		// by this, we ensure a representative peer may only represent the rule it was inferred from
+		// and uniqueness of representative peers.
+		// (another policy in another namespace, may have a rule with same podSelector, but the namespace will be different-
+		// so a different representative peer will be generated)
+		nsLabelSelector = &metav1.LabelSelector{MatchLabels: defaultNamespaceLabelsMap(podNs)}
```
adisos, line 630: verify here that `podNs` is not empty

pkg/netpol/eval/internal/k8s/peer.go
shireenf-ibm: it is never used for representative peer.

adisos: the code will be more robust if valid for this case as well, or if issue an error in that case as not supported for this function.

pkg/netpol/eval/resources.go
```diff
@@ -574,48 +585,76 @@ func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace st
...
-	}
-	// ingress-controller will be treated as a real pod, may be added to podsMap
 	pe.podsMap[podStr] = newPod
 	return &k8s.WorkloadPeer{Pod: newPod}, nil
 }
+
+// addRepresentativePod adds a new representative pod to the policy-engine (to pe.representativePeersMap).
+// if the given namespace string (podNs) is not empty (i.e. a real (policy's) namespace name), it will be assigned to the pod's Namespace;
+// and the "namespace name" requirement of the representative pod will be stored in its RepresentativeNsLabelSelector field.
+// if the representative pod will have no namespace (will not add a representative namespace to the policy-engine).
```
adisos, line 616:
```suggestion
// Otherwise, the representative pod will have no namespace (will not add a representative namespace to the policy-engine).
```

pkg/netpol/eval/resources.go
```diff
@@ -574,48 +585,76 @@ func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace st
...
+		// all representative pods are having same name since this name is used only to indicate that this Fake Pod is representative;
+		// this name is not used for storing it in the policy-engine/ comparing with other peers/ or representing it.
+		Name:                           k8s.RepresentativePodName,
+		Namespace:                      podNs,
+		FakePod:                        true,
+		RepresentativePodLabelSelector: objSelectors.PodSelector,
+		RepresentativeNsLabelSelector:  nsLabelSelector,
+	}
+	//  compute a unique string from the label selectors to be used as the map key
+	// note that nsLabelSelector might not be nil
```
adisos, line 642:
```suggestion
	// note that nsLabelSelector will not be nil
```

pkg/netpol/eval/resources.go
shireenf-ibm: this func is not relevant to exposure-analysis; was in the file from an old issue, the word representative here is not related to the "RepresentativePeer" that we use in exposure-analysis.

If I remember correctly; it was added for diff branch (as an owner might have more than one replica/ pod) and in the second dir one of these replicas might be deleted, so the `podOwnersToRepresentativePodMap` of policy-engine has to be updated.

I think, the `pe.podOwnersToRepresentativePodMap` is a policy-engine member from the initial versions of the netpol-analyzer

pkg/netpol/eval/resources.go
```diff
@@ -574,48 +585,76 @@ func (pe *PolicyEngine) GetSelectedPeers(selectors labels.Selector, namespace st
...
+	// note that nsLabelSelector will not be nil
+	nsKey, err := k8s.VariantFromLabelsSelector(nsLabelSelector)
+	if err != nil {
+		return err
+	}
+	podKey, err := k8s.VariantFromLabelsSelector(objSelectors.PodSelector)
+	if err != nil {
+		return err
+	}
+	keyStrFromLabels := nsKey + "/" + podKey
```
adisos, line 651: what if one policy has nil ns selector and pod selecotr {app: foo}, within namespace N1 [N1 is defined with label {x: xval}],
and another policy has ns selector {x: xval} and pod selecotr {app: foo} ?

pkg/netpol/eval/resources.go
adisos: right

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,116 @@
...
+	"encoding/hex"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+)
+
+// this file handles checking matching of representative peers' selectors' requirements
+// with rule selectors from policies
+// since a representative peer is a peer inferred from a network-policy rule
```
adisos, line 20:
```suggestion
// since a representative peer is a peer inferred from a network-policy rule
// for example, if a policy that captures a real workload `A` has a rule permitting egress connection, with selectors that match a representative peer `B`, 
// we need to capture this match, and thus infer the exposure connectivity from `A` to `B`
```

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,119 @@
...
+// for example, if a policy that captures a real workload `A` has a rule permitting egress connection,
+// with selectors that match a representative peer `B`,
```
adisos, line 22:
```suggestion
// SelectorsFullMatch is given two labelSelector objects, one from a policy rule and the other from a representative pod,
```

shireenf-ibm: this is not clear to me, applying it made no sense, so reverted

adisos: the line starts with "for example" should have been before the function description and not above it

shireenf-ibm: done

pkg/netpol/eval/check.go
```diff
@@ -280,63 +279,52 @@ func (pe *PolicyEngine) allallowedXgressConnections(src, dst k8s.Peer, isIngress
...
+			// update its ingress entire cluster connection relying on policy data
+			dst.GetPeerPod().UpdatePodXgressExposureToEntireClusterData(policy.IngressPolicyExposure.ClusterWideExposure, isIngress)
+		}
+		// get ingress allowed conns between src and dst
 		switch {
-		case policy.IngressGeneralConns.AllDestinationsConns.AllowAll:
-			return policy.IngressGeneralConns.AllDestinationsConns, nil
-		case policy.IngressGeneralConns.EntireClusterConns.AllowAll && src.PeerType() == k8s.PodType:
-			return policy.IngressGeneralConns.EntireClusterConns, nil
+		case policy.IngressPolicyExposure.ExternalExposure.AllowAll:
```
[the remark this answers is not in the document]
shireenf-ibm: although the first two cases might have a `true` value only when exposure analysis is on;
the connection sets of `policy.IngressPolicyExposure` are initiated to an empty ConnSet (AllowAll == false) also when exposure analysis is off.
so the results will not be wrong/ affected when the exposure analysis is inactive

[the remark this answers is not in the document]
adisos: right, this comment is more for readability reasons.. I find it a bit confusing that one section in this func checks for exposure analysis activated, and then the other section mixes both cases.

pkg/netpol/eval/exposure.go
shireenf-ibm: not really required, removed it

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -141,8 +143,8 @@ func (np *NetworkPolicy) ruleConnections(rulePorts []netv1.NetworkPolicyPort, ds
 	return res, nil
 }
 
-// isRepresentativePod determines if the peer's source is representativePeer; i.e. its pod fake and has RepresentativePodName
-func isRepresentativePod(peer Peer) bool {
+// hasRepresentativePod determines if the peer's source is representativePeer; i.e. its pod fake and has RepresentativePodName
+func hasRepresentativePod(peer Peer) bool {
```
adisos, line 147: rename `hasRepresentativePod` to `isPeerRepresentative` ?

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -200,27 +202,42 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
...
 			peerMatchesPodSelector := false
 			peerMatchesNamespaceSelector := false
+			var err error
+			var selector labels.Selector
 			if rulePeers[i].NamespaceSelector == nil {
 				peerMatchesNamespaceSelector = (np.ObjectMeta.Namespace == peer.GetPeerPod().Namespace)
 			} else {
-				selector, err := np.parseNetpolLabelSelector(rulePeers[i].NamespaceSelector)
+				peerNamespace := peer.GetPeerNamespace()
+				// checking if a selector matches labels by peer type; since representative peers selectors may need special handling
```
adisos, line 211:
```suggestion
				// for exposure analysis, use relevant func to check if representative peer is matched by rule's selector 
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -200,27 +202,42 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
...
+				if hasRepresentativePod(peer) {
+					// representative peer is inferred from a rule with its/same labelSelector.
+					// note that if the namespaceSelector in the rule is nil, we don't get here,
+					// since that means the peer is in same namespace of the policy, and this was checked above
+					peerMatchesNamespaceSelector, err = SelectorsFullMatch(rulePeers[i].NamespaceSelector, peer.GetPeerPod().RepresentativeNsLabelSelector)
+				} else {
```
adisos, line 217:
```suggestion
				} else { // for real peer just check if the selector matches the peer's labels 
```

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -200,27 +202,42 @@ func (np *NetworkPolicy) ruleSelectsPeer(rulePeers []netv1.NetworkPolicyPeer, pe
...
-				peerMatchesNamespaceSelector = selector.Matches(labels.Set(peerNamespace.Labels))
 			}
 			if !peerMatchesNamespaceSelector {
 				continue // skip to next peerObj
 			}
 			if rulePeers[i].PodSelector == nil {
 				peerMatchesPodSelector = true
 			} else {
-				selector, err := np.parseNetpolLabelSelector(rulePeers[i].PodSelector)
+				// checking if a selector matches labels by peer type; since representative peers selectors may need special handling
```
adisos, line 231: can capture this common code in a separate function so it is not duplicated?

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+// 2. if the rule's selector is empty (matches all pods/namespaces)
+// 3. if the requirements of both rules are equal (same)
+// i.e. checks if the selector from the policy rule fully matches the selector of the representative peer
+// note that : for non-empty selectors that partly match/ contained in each other the func returns false
+func SelectorsFullMatch(ruleSelector, repSelector *v1.LabelSelector) (bool, error) {
+	if ruleSelector == repSelector { // both label selectors point to same reference
+		return true, nil
+	}
+	// otherwise, check requirements equality
+	selector1, err := v1.LabelSelectorAsSelector(ruleSelector)
```
adisos, line 38: is it required to add other (different) representative peers as captured by empty rules? 
what is going to be the final outcome? exposure expressed for all  representative peers for every empty rule? 
shouldn't refine to requiring that also  repSelector is empty in this case?

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+// with selectors that match a representative peer `B`, we need to capture this match,
+// and thus infer the exposure connectivity from `A` to `B`.
+// returns true if:
+// 1. both selectors point to same reference (rule and its matching representative pod/ns)
+// 2. if the rule's selector is empty (matches all pods/namespaces)
+// 3. if the requirements of both rules are equal (same)
+// i.e. checks if the selector from the policy rule fully matches the selector of the representative peer
+// note that : for non-empty selectors that partly match/ contained in each other the func returns false
+func SelectorsFullMatch(ruleSelector, repSelector *v1.LabelSelector) (bool, error) {
+	if ruleSelector == repSelector { // both label selectors point to same reference
```
adisos, line 34: rename `selector1` to `ruleSelectorConverted`

adisos:
is there a convention for connlist tests with exposure analysis?
how can one identify which tests are relevant to exposure analysis?

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+		return true, nil
+	}
+	// otherwise, check requirements equality
+	selector1, err := v1.LabelSelectorAsSelector(ruleSelector)
+	if err != nil {
+		return false, err
+	}
```
adisos, line 41: rename `selector2` to `repSelectorConverted`

shireenf-ibm:
> is there a convention for connlist tests with exposure analysis? how can one identify which tests are relevant to exposure analysis?

I tried to start all dirs of tests with `test_exposure` , you may also see the tests in `connlist_test.go` which run with the `exposureFlag`. all output files of tests that run with exposure-analysis (on) will start with `exposure_` (determined this in code )

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+		return true, nil
+	}
+	selector2, err := v1.LabelSelectorAsSelector(repSelector)
+	if err != nil {
+		return false, err
+	}
+	// Requirements() returns sorted by key list;
+	// since `v1.LabelSelectorAsSelector` builds a new `labels.selector` by using `Requirements.Add` which sorts the requirements slice by key.
+	// and `Requirements()` returns this sorted list
+	// links:
```
adisos, line 52: rename requirements1,requirements2, to ruleRequirements and repRequirements

adisos:
> > is there a convention for connlist tests with exposure analysis? how can one identify which tests are relevant to exposure analysis?
> 
> I tried to start all dirs of tests with `test_exposure` , you may also see the tests in `connlist_test.go` which run with the `exposureFlag`. all output files of tests that run with exposure-analysis (on) will start with `exposure_` (determined this in code )

can you be consistent also with test dirs, so that they all have a common prefix? some start with 
"test_new_namespace_conn_and_entire_cluster_with_matching_pod" , some with "test_exposure_", some with "test_egress_exposure", and maybe others...

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+		if newStr2 := handleRequirementWithInOpAndSingleValue(requirements2[i]); newStr2 != "" {
+			str2 = newStr2
+		}
+		if str1 != str2 {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
```
adisos, line 88: where does this str come from? better build a relevant selector object and use its string() function.

adisos:
Please add a short summary of the implementation flow in the issue description.

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -0,0 +1,120 @@
...
+// handleRequirementWithInOpAndSingleValue returns a <key>=<val> string format if the input Requirement is with In operator and single value
+func handleRequirementWithInOpAndSingleValue(req labels.Requirement) string {
+	if req.Operator() == selection.In && len(req.Values()) == 1 {
+		return req.Key() + "=" + req.Values().List()[0]
+	}
+	return ""
+}
```
adisos, line 95: the name `VariantFromLabelsSelector` is misleading.. consider renaming to UniqueKeyFromLabelsSelector

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: - the case is when the rule is empty, its connection (port) is applied to all peers in the cluster (real + representative); by default this connection will appear in the output for real peers, so I thought it should appear also for the representative peer.
- we don't have representative peer with empty selectors; this connection will also appear for the <entire-cluster> node.

adisos: will open a separate task in the issue to consider this:
- for a representative peer, exposure analysis propagates connections  from exposure of entire cluster
- but,  it does not propagate connections from exposure to another rep-peer with match-expression that "contain" this peer

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: > please add a comment and example for the current reason that true is returned for an empty selector.

done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -53,30 +60,30 @@ func SelectorsFullMatch(ruleSelector, repSelector *v1.LabelSelector) (bool, erro
...
-		// requirements1[i].Equal(requirements2[i]) returns false if the values are not in same order
+	for i := range ruleRequirements {
+		// ruleRequirements[i].Equal(repRequirements[i]) returns false if the values are not in same order
 		// (stringslices.Equal returns true only if two slices have same length and same order of items)
 		// however, `Requirement.String` sorts the values using `safeSort`, so will compare by string
 		// link to Requirement.String() :
 		// https://github.com/kubernetes/apimachinery/blob/d7e1c5311169d5ece2db0ae0118066859aa6f7d8/pkg/labels/selector.go#L310
-		str1 := requirements1[i].String()
-		str2 := requirements2[i].String()
+		str1 := ruleRequirements[i].String()
```
adisos, line 74: change also str1, str2, to follow same convention

pkg/netpol/eval/internal/k8s/representative_selectors.go
```diff
@@ -35,14 +35,21 @@ func SelectorsFullMatch(ruleSelector, repSelector *v1.LabelSelector) (bool, erro
...
 	}
 	// otherwise, check requirements equality
-	selector1, err := v1.LabelSelectorAsSelector(ruleSelector)
+	ruleSelectorConverted, err := v1.LabelSelectorAsSelector(ruleSelector)
 	if err != nil {
 		return false, err
 	}
-	if selector1.Empty() { // empty rule matches everything
+	if ruleSelectorConverted.Empty() { // empty rule matches everything
+		// returning true, so we can capture the connection with this rule and add it to the exposure connection of the representative peer.
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/check.go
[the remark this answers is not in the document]
shireenf-ibm: in [commit](https://github.com/np-guard/netpol-analyzer/pull/377/commits/a3f434c7b4af35be0f6699d56bca01e0931e868b) , split the func into two, one for returning allowed conns between peers, and the other for updating the pod's exposure (the second used only with active exposure-analysis)

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: updated comment

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: updated the comment maybe it is more clear now.
k8s package doesn't have the exposure flag, but has the pods instances; so the used condition is equivelent to checking if exposure analysis is active

pkg/netpol/eval/internal/k8s/peer.go
shireenf-ibm: I've widen the case of fake-pod for "representative-peer" too.
it is simply not used since we don't care of the String representation of a representative peer; we just need its labels. 
so I don't think should raise an error too.

pkg/netpol/eval/internal/k8s/pod.go
shireenf-ibm: will not - updated

shireenf-ibm: will not.. updated

pkg/netpol/eval/internal/k8s/netpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/resources.go
shireenf-ibm: yes, done

pkg/netpol/eval/internal/k8s/peer.go
adisos: ok, better add a comment that it is not expected to be used for  representative peer.

pkg/netpol/eval/resources.go
shireenf-ibm: - if there is a real pod in `N1` then the representative peers inferred from both policy rules will be removed (refined) and not appear in the output.
- if there is no real pod in the ns, then we will see 2 representative peers; a representative peer in `N1` and another representative-peer in any-ns with {x:val} >> since we refine rep-peers matching pods; not matching namespaces.
- added tests

pkg/netpol/eval/internal/k8s/peer.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: as described in the comments `string()` func of selector is not helpful for our case.
we loop sorted selector's requirements and compare strings of the requirements. (if not all requirements strings equal, then selectors not equal).

the use of `handleRequirementWithInOpAndSingleValue` is only for a single case of a single requirement. what it does actually:
if the requirement applies the condition (In operator with single value) it replaces its string to be <key> = <val> instead of <key> in <val>.
so we ensure two requirements {<key> in <val>} and {<key> = <val>} are treated the same.

renamed the func to `replaceStringOfRequirementWithInOpAndSingleValue`

is this more clear?

(creating a new selector with all other requirements too may be time costing; and we can't predict if any of the requirements will have{<key> in <val>} )

shireenf-ibm:
> Please add a short summary of the implementation flow in the issue description.

done

shireenf-ibm:
> > > is there a convention for connlist tests with exposure analysis? how can one identify which tests are relevant to exposure analysis?
> > 
> > 
> > I tried to start all dirs of tests with `test_exposure` , you may also see the tests in `connlist_test.go` which run with the `exposureFlag`. all output files of tests that run with exposure-analysis (on) will start with `exposure_` (determined this in code )
> 
> can you be consistent also with test dirs, so that they all have a common prefix? some start with "test_new_namespace_conn_and_entire_cluster_with_matching_pod" , some with "test_exposure_", some with "test_egress_exposure", and maybe others...

done , described the changes in the PR's description above too

pkg/netpol/eval/internal/k8s/representative_selectors.go
adisos: please generate a new req object , to get the string from it

pkg/netpol/eval/resources.go
adisos: meant to the branch `if nsLabelSelector == nil `

pkg/netpol/eval/internal/k8s/representative_selectors.go
shireenf-ibm: done

pkg/netpol/eval/resources.go
shireenf-ibm: done

adisos (approved):

Issue #30, opened 2022-10-27
Introduce support for (cluster scoped) AdminNetworkPolicy API.


elevran:
k8s is adding support for cluster scoped NetworkPolicy objects see [here](https://github.com/kubernetes/enhancements/issues/2091) and [here](https://github.com/kubernetes-sigs/network-policy-api/pull/30). Note that there are both `Admin` and `Baseline` objects to be accounted for.

Explore how to enable these in the policy engine.

zivnevo:
Antrea [now support](https://antrea.io/docs/v1.13.0/docs/admin-network-policy/) `AdminNetworkPolicies` and `BaselineAdminNetworkPolicy`.

zivnevo:
OpenShift native CNI is also going to support these. See [here](https://github.com/ovn-org/ovn-kubernetes/blob/master/docs/networkpolicies/admin-network-policy.md)

PR #380, opened 2024-07-08
W.I.P Support admin netpolicy

shireenf-ibm:
#30


shireenf-ibm:
hi @adisos. this is the beginning of supporting the ANP with first simple tests

i still have TODO's , like:
- some @todo comments in code (the code is not final for review)
- extend the egress rulePeers (AdminNetworkPolicyEgressPeer)  to support also Nodes and Networks fields.
- add more tests with multiple ANPs and multiple rules (and actions) - (this will help me check the correctness of the computations of allowed conns)
- add tests which involves ANPs and NPs 
- add bad path tests 
- support also BANP

shireenf-ibm:
will close for a while , not ready for review - writing some critical changes

shireenf-ibm:
hi @adisos , @tanyaveksler ,
while reviewing : please note that the branch contains some `@todo` / `TBD` comments;
those are intended to be resolved in further branches (that will be created from this branch according to this [tasks list](https://github.com/np-guard/netpol-analyzer/pull/380#issuecomment-2214896915) ).

pkg/internal/netpolerrors/netpol_errors.go
```diff
@@ -126,3 +136,19 @@ const colonSep = ": "
 func ConcatErrors(err1, err2 string) string {
 	return err1 + colonSep + err2
 }
+
+// SamePriorityErr returns the error message of a priority appears more than once in different admin-network-policies
+func SamePriorityErr(name1, name2 string) string {
+	return "Admin Network Policies: " + name1 + " and " + name2 + " has same priority;" +
```
tanyaveksler, line 142:
```suggestion
	return "Admin Network Policies: " + name1 + " and " + name2 + " have same priority;" +
```

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+	if anpSubject.Namespaces != nil {
+		return anp.doesNamespacesFieldMatchPeer(anpSubject.Namespaces, p)
+	}
+	// else: Subject.Pods is not empty (Subject.Pods contains both NamespaceSelector and PodSelector)
+	return anp.doesPodsFieldMatchPeer(anpSubject.Pods, p)
+}
+
+// onlyOnePeersFieldIsSet returns if exactly one of the given objects is set (not nil)
+func onlyOnePeersFieldIsSet(namespacesField *metav1.LabelSelector, podsField *apisv1a.NamespacedPod) bool {
+	return !((namespacesField == nil && podsField == nil) || (namespacesField != nil && podsField != nil))
```
tanyaveksler, line 67:
```suggestion
	return (namespacesField == nil) != (podsField == nil)
```

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		if err != nil {
+			return false, err
+		}
+		if fieldMatch {
+			return true, nil
+		}
+	}
+	return false, nil
+}
```
tanyaveksler, line 241: The two functions are nearly identical, can reuse the same code?

pkg/netpol/eval/eval_test.go
```diff
@@ -1061,7 +1061,8 @@ func setResourcesFromDir(pe *PolicyEngine, path string, netpolLimit ...int) erro
 	var netpols = []*netv1.NetworkPolicy{}
 	var pods = []*v1.Pod{}
 	var ns = []*v1.Namespace{}
-	for _, obj := range objectsList {
+	for i := range objectsList {
+		obj := objectsList[i]
```
tanyaveksler, line 1065: Why did you do this change here and in other places?

shireenf-ibm: to get the linter happier :)
when updated the parser with the new `ANP` object, it raised a linting err called `rangeValCopy` which  detects loops that copy big objects during each iteration; resolved by using index access (as linter suggested)

pkg/netpol/internal/common/portset.go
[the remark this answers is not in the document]
shireenf-ibm: true, fixed

tanyaveksler (dismissed):
Good work, @shireenf-ibm !
I had a few comments which you've already handled.

shireenf-ibm:
hi @tanyaveksler , @adisos 
I have opened two new PRs from this PR.
-  #402 : sorts all the ANPs once, and storing them as sorted slice in the policy-engine (instead of repeatedly sorting ANPs that select src/dst when computing allowed conns only) 

- #403 : added support to baseline-admin-networkpolicy (BANP); 
this PR contains also multiple updates to code of computing allowed connections between two peers (in `check.go` and `policy_connections.go`, inter alia, resolving some @todo, which were related to type of allowed connections.
it also contains doc updates explaining more broadly steps and funcs of computing allowed-connections

adisos (changes requested):
looks good, added few comments, did not review all files yet

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+		err = res.UpdateWithRuleConns(ruleConns, rule.Action)
+		if err != nil {
+			return nil, anp.anpErr(fmt.Sprintf(ruleErrTitle, rule.Name), err.Error())
+		}
+	}
+	return res, nil
+}
+
+const (
+	minPriority = 0
```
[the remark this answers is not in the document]
shireenf-ibm: yes, [here](https://network-policy-api.sigs.k8s.io/api-overview/#adminnetworkpolicy-priorities)

> The Priority field in the ANP spec is defined as an integer value within the range 0 to 1000

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,87 @@
...
+	apisv1a "sigs.k8s.io/network-policy-api/apis/v1alpha1"
+
+	"github.com/np-guard/netpol-analyzer/pkg/internal/netpolerrors"
+	"github.com/np-guard/netpol-analyzer/pkg/netpol/internal/common"
+)
+
+// PolicyConnections : stores connections collected mainly from admin-network-policies and adjusted with allowed netpols conns
+// or default conns
+type PolicyConnections struct {
+	AllowedConns *common.ConnectionSet
```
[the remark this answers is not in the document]
shireenf-ibm: resolved in #403

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// anpErr returns string format of an admin netpol err
+func (anp *AdminNetworkPolicy) anpErr(title, description string) error {
+	return fmt.Errorf("admin network policy %q %s: %s", anp.Name, title, description)
+}
+
+// GetIngressPolicyConns returns the connections from the ingress rules selecting the src in spec of the adminNetworkPolicy
+//
+//nolint:dupl // this loops Ingress spec - different types
```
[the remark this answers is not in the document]
shireenf-ibm: I am trying to see if generics help with this and more funcs in #403

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+	podSelector, err := anp.parseAdminNetpolLabelSelector(&pods.PodSelector)
+	if err != nil {
+		return false, err
+	}
+	return nsSelector.Matches(labels.Set(peer.GetPeerNamespace().Labels)) && podSelector.Matches(labels.Set(peer.GetPeerPod().Labels)), nil
+}
+
+// parseAdminNetpolLabelSelector returns a selector of type labels.selector from a LabelSelector in the policy.
+// an error with the admin network policy details returned if fails to convert the selector.
+func (anp *AdminNetworkPolicy) parseAdminNetpolLabelSelector(selector *metav1.LabelSelector) (labels.Selector, error) {
```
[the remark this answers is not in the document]
shireenf-ibm: this function is eliminated in #403

pkg/netpol/eval/resources.go
```diff
@@ -422,6 +436,21 @@ func (pe *PolicyEngine) insertNetworkPolicy(np *netv1.NetworkPolicy) error {
 	return err
 }
 
+func (pe *PolicyEngine) insertAdminNetworkPolicy(anp *apisv1a.AdminNetworkPolicy) error {
+	// @TBD : currently disabling exposure-analysis when there are admin-network-policies in the input resources
+	if pe.exposureAnalysisFlag {
+		return errors.New(netpolerrors.ExposureAnalysisDisabledWithANPs)
+	}
+	if anp.Name == "" {
```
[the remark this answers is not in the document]
shireenf-ibm: no, I have no idea why; names are used to store the objects, and an object without a name would cause wrong results.
maybe, since our tests was always from `dir` the `yaml` correctness was assumed to be.
I had the need to check the names after the missing names of objects in new tests added by Tanya caused unexpected results.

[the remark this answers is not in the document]
adisos: can try adding a badpath test with policy without name?
where in the flow does it cause an issue if admin policy has no name? if it is common also for regular netpol, can try debug that run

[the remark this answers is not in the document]
shireenf-ibm: actually,
tests that failed were tests with more than one ANP when all policies without names, so the policy-engine acted as if only one policy was provided only; the problem was not the empty name string but the uniqueness of the names.

1. when we have a single policy, we will not have wrong results (it will be stored in the map as `{"":&anp}` (with empty string as its name).
but when we have 2 policies with empty names, only one will be stored in the map with this empty string key and then we'll have unexpected results
2. according to [this](https://kubernetes.io/docs/concepts/overview/working-with-objects/names#names) names must be unique 

so maybe the next check `if there is already a policy with this name` is enough?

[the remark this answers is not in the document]
adisos: yes, so better check for unique names (namespace + name) for both regular netpols and admin netpols

go.mod
```diff
@@ -63,5 +63,6 @@ require (
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/kustomize/api v0.13.5-0.20230601165947-6ce0bf390ce3 // indirect
 	sigs.k8s.io/kustomize/kyaml v0.14.3-0.20230601165947-6ce0bf390ce3 // indirect
+	sigs.k8s.io/network-policy-api v0.1.5
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/manifests/parser/k8sobj.go
```diff
@@ -199,7 +206,8 @@ func FilterObjectsList(allObjects []K8sObject, podNames []types.NamespacedName)
 		nsMap[podNames[i].Namespace] = true
 	}
 	res := make([]K8sObject, 0)
-	for _, obj := range allObjects {
+	for i := range allObjects {
+		obj := allObjects[i]
 		switch obj.Kind {
```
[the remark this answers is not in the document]
shireenf-ibm: this func is related only to `eval` command, I prefer to update if needed in the relevant PR that will support the eval command.
currently, seemed to me that it filters objects that have namespaces only, but ANPs are cluster-wide and are not scoped by any namespace.

pkg/netpol/eval/resources.go
```diff
@@ -107,7 +110,8 @@ func (pe *PolicyEngine) AddObjectsForExposureAnalysis(objects []parser.K8sObject
 }
 
 func splitPoliciesAndNamespacesAndOtherObjects(objects []parser.K8sObject) (policiesAndNs, others []parser.K8sObject) {
-	for _, obj := range objects {
+	for i := range objects {
+		obj := objects[i]
```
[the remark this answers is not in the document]
shireenf-ibm: currently no, this is called for exposure-analysis only.
and for now, exposure-analysis is diasbled when there are admin-netpols in the resources.
if we wish to enable exposure-analysis with projects that have ANPs too, i will add a task and handle it in a seperate PR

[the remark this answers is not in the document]
adisos: ok, please add a task and a todo comment here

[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/internal/common/connectionset.go
```diff
@@ -119,6 +120,42 @@ func (conn *ConnectionSet) Union(other *ConnectionSet) {
...
+		conn.AllowedProtocols = map[v1.Protocol]*PortSet{}
+		return
+	}
+	if conn.AllowAll {
+		conn.AllowAll = false // we are about to subtract something
+		conn.addAllConns()
+	}
+	for protocol, ports := range conn.AllowedProtocols {
+		if otherPorts, ok := other.AllowedProtocols[protocol]; ok {
+			if ports.Equal(otherPorts) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/internal/common/connectionset.go
```diff
@@ -119,6 +120,42 @@ func (conn *ConnectionSet) Union(other *ConnectionSet) {
...
+				delete(conn.AllowedProtocols, protocol)
+			} else {
+				ports.subtract(otherPorts)
+				if conn.AllowedProtocols[protocol].IsEmpty() {
```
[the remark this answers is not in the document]
shireenf-ibm: true, done

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,297 @@
...
+	if isIngress {
+		return len(anp.Spec.Ingress) > 0
+	}
+	return len(anp.Spec.Egress) > 0
+}
+
+// subjectSelectsPeer returns if the subject of the adminNetworkPolicy selects the given peer
+func (anp *AdminNetworkPolicy) subjectSelectsPeer(p Peer) (bool, error) {
+	anpSubject := anp.Spec.Subject
+	if !onlyOnePeersFieldIsSet(anpSubject.Namespaces, anpSubject.Pods) {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/resources.go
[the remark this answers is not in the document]

pkg/manifests/parser/k8sobj.go
```diff
@@ -44,8 +47,9 @@ type K8sObject struct {
 	// namespace object
 	Namespace *v1.Namespace
 
-	// netpol object
-	NetworkPolicy *netv1.NetworkPolicy
+	// netpol objects
+	NetworkPolicy      *netv1.NetworkPolicy
+	AdminNetworkPolicy *apisv1a.AdminNetworkPolicy
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/pod.go
```diff
@@ -277,16 +277,15 @@ func (pod *Pod) PodExposedTCPConnections() *common.ConnectionSet {
 	return res
 }
 
-// ConvertPodNamedPort returns the ContainerPort number that matches the named port
-// if there is no match, returns -1
-// namedPort is unique within the pod
-func (pod *Pod) ConvertPodNamedPort(namedPort string) int32 {
+// ConvertPodNamedPort returns the ContainerPort's protocol and number that matches the named port
+// if there is no match, returns empty string for protocol and -1 for number
+func (pod *Pod) ConvertPodNamedPort(namedPort string) (protocol corev1.Protocol, portNum int32) {
```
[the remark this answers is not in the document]
shireenf-ibm: seems to me needs a fix,
 I ran an example where  `pod-a` has following named-port : 
```
- name: newport
   containerPort: 8956
   protocol: SCTP
```

and a netpol capturing this `pod-a` with an ingress rule from any other pod with following ports:
```
    ports:
    - port: newport
      protocol: UDP
```

in the result got a connection to the pod with `UDP 8956` 
`helloworld/new-pod[Deployment] => helloworld/pod-a[Deployment] : UDP 8956`

i believe we shouldn't see such a connection as there is no match between the rule's port and the deployment's port.

opening a new issue with the example #405

pkg/netpol/eval/internal/k8s/adminnetpol.go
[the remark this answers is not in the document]
shireenf-ibm: currently tried to see if generics helps to avoid code duplication with  `ingressRuleSelectsPeer` and `egressRuleSelectsPeer` func, but unfortunately it didn't work because of a Go restriction on generic types. 

> "The Go compiler does not support accessing a struct field x.f where x is of type parameter type even if all types in the type parameter’s type set have a field f. We may remove this restriction in a future release."

(taken from https://tip.golang.org/doc/go1.18#generics and still relevant to go 1.21)

i have committed [here](https://github.com/np-guard/netpol-analyzer/pull/403/commits/c1bcdcd99787f151c337859462855a9ee1f54065) (in #403) an explanation "why could not use generics in this case)
and still trying to see if there are other cases where generics can be used

shireenf-ibm:
hi @tanyaveksler , @adisos:

new commits for your review:
- addressed Adi's comments above
- merged two branches into this : #402 & #403 
those two branches included multiple code and doc enhancements to current PR's code (+ resolving some important @todo-s), so better be reviewed once

tanyaveksler (commented):
@shireenf-ibm, the code looks good, great work!
I reviewed 4 commits since the merge with `tests_for_admin_netpolicy` branch.
I have a few comments/suggestions.

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -69,16 +67,52 @@ func (pc *PolicyConnections) UpdateWithRuleConns(ruleConns *common.ConnectionSet
...
+	passAllowCopy := pc.PassConns.Copy() // using a copy since Intersection changes the object, but we want to keep also
+	// non-intersected conns
+	passAllowCopy.Intersection(otherConns.AllowedConns) // pass conns to be allowed
+	passDenyCopy := pc.PassConns.Copy()
+	passDenyCopy.Intersection(otherConns.DeniedConns) // pass conns to be denied
+	// update current's allowed and denied conns with:
+	// 1. determined pass conns
+	// 2. with traffic that had no match in ANP (or higher priority policies)
+	pc.AllowedConns.Union(passAllowCopy)
+	pc.AllowedConns.Union(otherConns.AllowedConns)
```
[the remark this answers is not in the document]
shireenf-ibm: * this line `pc.AllowedConns.Union(otherConns.AllowedConns)` is for adding "extra" allowed traffic that was not mentioned at all in ANP (or higher priority policies)
lets assume `otherConns.AllowedConns` is coming from a BaselineAdminNetworkPolicy and `pc.AllowedConns` contains allowed conns from an AdminNetworkPolicy; (an example without regular netpols)
if `otherConns.AllowedConns` contains a traffic (ports) that was not mentioned in the ANP at all (not in pass nor allow nor deny), then we expect to see it also as allowed connections in the connectivity report sinc the BANP allows it.
so this row is responsible for adding the allowed-conns that were not mentioned in a higher-level policies.

I have added a test with similar case `tests\anp_with_banp_new_test` (in this example ANP has only one pass rule with TCP-80, and BANP allows TCP80-81 and denies all other conns; expected result is to see TCP80-81 (even though TCP 81 and other conns are not mentioned in the ANP))
- generated this test case in policy assistant too and got expected conns (TCP80-81 area allowed)

however it seems that the previous line `pc.AllowedConns.Union(passAllowCopy)` is redundant (passAllowCopy is contained in other.AllowedConns which after subtract lines above contians only passed and new "extra" conns -  rearranging this func)

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -69,16 +67,52 @@ func (pc *PolicyConnections) UpdateWithRuleConns(ruleConns *common.ConnectionSet
...
+	pc.DeniedConns.Union(passDenyCopy)
+	pc.DeniedConns.Union(otherConns.DeniedConns)
```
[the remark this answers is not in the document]
shireenf-ibm: no, `otherConns.DeniedConns` contains passed traffic and new traffic that was not declared in any of the ANPs;
updated the func with comments, more explanation [here](https://github.com/np-guard/netpol-analyzer/pull/380#discussion_r1761119598)

tanyaveksler (commented):
I have just a few small comments.

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -94,19 +94,27 @@ func (pc *PolicyConnections) CollectConnsFromLowerPolicyType(otherConns *PolicyC
 	otherConns.AllowedConns.Subtract(pc.DeniedConns)
```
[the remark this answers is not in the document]
shireenf-ibm: no, this func will "convert" `pc.PassConn` to allowed or denied based on the otherConns.AllowedConns and otherConns.DeniedConns.

the idea from this func is that the current `pc` contains the collected final conns from all the `ANPs` 
which means: 
- `pc.AllowedConns` are definitely allowed
-` pc.DeniedConns` are definitely denied 

and pc.PassedConns will be determined by the new `otherConns` object which is coming from a PolicyType with lower priority than ANP (i.e. a NetworkPolicy or a BANP);
so after subtracting the finite allowed and denied conns of `pc` from `otherConns` (lines 94-94) ; otherConns will contain only new allowed and denied conns. 
so we won't subtract the `otherConns.AllowedConns` from the pc.PassConn;
neither subtracting `otherConns.DeniedConns` from `pc.PassConns` 
vice versa the `otherConns.AllowedConns` will be added to (union) the `pc.allowedConns`
and the `otherConns.DeniedConns` will be added to the `pc.DeniedConns` 
hopefully that by this all `pc.PassConns` are covered

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -94,19 +94,27 @@ func (pc *PolicyConnections) CollectConnsFromLowerPolicyType(otherConns *PolicyC
...
 	otherConns.DeniedConns.Subtract(pc.AllowedConns)
```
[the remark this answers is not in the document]
shireenf-ibm: explained [above](https://github.com/np-guard/netpol-analyzer/pull/380#discussion_r1765032947)

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -94,19 +94,27 @@ func (pc *PolicyConnections) CollectConnsFromLowerPolicyType(otherConns *PolicyC
...
+	// 2. traffic that had no match in ANP (or higher priority policies)
+	// so we can update current denied conns with otherConns.DeniedConns
+	pc.DeniedConns.Union(otherConns.DeniedConns)
+
+	// in order to update pc.PassConns we need:
+	// to find intersection of current pass connections with otherConns's allowedConns and deniedConns
 	passAllowCopy := pc.PassConns.Copy() // using a copy since Intersection changes the object, but we want to keep also
 	// non-intersected conns
 	passAllowCopy.Intersection(otherConns.AllowedConns) // pass conns to be allowed
+	// pc.AllowedConns.Union(passAllowCopy) - redundant since passAllowCopy is contained in otherConns.AllowedConns, already updated
```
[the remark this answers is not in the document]
shireenf-ibm: it is used at the end of the func (line 122-123); 
`passAllowCopy` and `passDenyCopy` are subtracted from `pc.PassConn` to have an indication if all Passed by ANPs conns are covered or not. (what is not coveered will finally be determined according to system-default which is allow-all)

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -94,19 +94,27 @@ func (pc *PolicyConnections) CollectConnsFromLowerPolicyType(otherConns *PolicyC
...
-	// 2. with traffic that had no match in ANP (or higher priority policies)
-	pc.AllowedConns.Union(passAllowCopy)
-	pc.AllowedConns.Union(otherConns.AllowedConns)
-	pc.DeniedConns.Union(passDenyCopy)
-	pc.DeniedConns.Union(otherConns.DeniedConns)
+	// pc.DeniedConns.Union(passDenyCopy) - redundant since passDenyCopy is contained in otherConns.DeniedConns
+
 	// subtract pass-deny and pass-allow from the current Pass conns;
 	// note that the updated pc conns may still have non-empty Pass connections (intersection with allow and deny are not full)
 	// - this will not affect evaluated netpols conns, as the allowed conns of netpols implicitly deny other conns.
```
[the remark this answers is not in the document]
shireenf-ibm: it is subtracted on line 122 (after this comment)

pkg/netpol/eval/check.go
```diff
@@ -257,12 +262,12 @@ func (pe *PolicyEngine) allowedXgressConnection(src, dst k8s.Peer, isIngress boo
 		if dst.PeerType() == k8s.IPBlockType {
 			return true, nil // all connections allowed - no restrictions on ingress to externalIP
 		}
```
tanyaveksler, line 264: This 'if is redundant - this check was moved to `getPoliciesSelectingPod`

pkg/netpol/eval/check.go
```diff
@@ -257,12 +262,12 @@ func (pe *PolicyEngine) allowedXgressConnection(src, dst k8s.Peer, isIngress boo
...
-		netpols, err = pe.getPoliciesSelectingPod(dst.(*k8s.PodPeer).Pod, netv1.PolicyTypeIngress)
+		netpols, err = pe.getPoliciesSelectingPod(dst, netv1.PolicyTypeIngress)
 	} else {
 		if src.PeerType() == k8s.IPBlockType {
 			return true, nil // all connections allowed - no restrictions on egress from externalIP
 		}
```
tanyaveksler, line 269: Same as above

tanyaveksler (commented):
I added small comment for removing some redundant code.

pkg/netpol/eval/check.go
shireenf-ibm: removed

pkg/netpol/eval/check.go
shireenf-ibm: done

tanyaveksler (dismissed):

shireenf-ibm:
hi @tanyaveksler , @adisos 
(a FYI comment)
while implementing support for `eval` command, I have thought of cases where the result that should be returned was not clear to me;
example case:

> if we have an ANP with subject selecting all-namespaces;
> an the ingress rule says: allow TCP 80 from ns:y
> but the egress rule says: deny TCP 80 to ns:x
> what would be the connection from y/pods->x/pods?

so I've generated new tests (for this case and other cases with "non-match" between ingress and egress rule); and ran them in the policy-assistant repo.

I am attaching the tests file , for each case in the file; added a short description and the actual-table that I've got from debugging the test in policy-assistant. (some results surprised me)
and I want to add these tests to this branch, to compare results and update functionality when needed.


(the file was added to `network-policy-api\cmd\policy-assistant\test\integration\shireen_test.go`)
[shireen_test.go](https://github.com/user-attachments/files/17190714/shireen_test.go.txt)

shireenf-ibm:
- in last [commit](https://github.com/np-guard/netpol-analyzer/pull/380/commits/793d3cb590622041a77a371c5b9bc5957f6b5d5f); 
I have added some new parsed tests and updated the `func getAdminPolicyConnFromEgressIngressConns` which computes the "intersection" between the ingress and egress when both capture connections between the `src` and `dst`.
 
- for each new test : I have added a description and the output table if we run this test with policy-assistant.
my expected result differs from `policy-assistant` output in two tests :
1. `ANP with unmatched ingress and egress connection #3`
2. `ANP with unmatched ingress and egress connection #14`

- my approach when ingress rules and egress rules capture both `src` and `dst` at the "layer" of ANP:
  * a connection is allowed iff both egress and ingress rules allow it ( i.e. if the connection appears as allowed on both directions, or allowed by one direction and not-captured by the other direction (not denied/passed by the other direction)))
  * a connection is denied if at least denied by one direction
  * a connection is passed if one direction passes it ( and was not denied by the other).
 
 ------ 
 later will update what happens in policy-assistant that causes different results from above;
 - currently from those tests seems to me that their approach for allow is : a connection is allowed if at least allowed by one direction

adisos:
> otherwise, a connection is passed if one direction passes it

not sure I understand.. why is entire connection "passed" ?can't it be passed in ingress, and allowed in egress?

shireenf-ibm:
> > otherwise, a connection is passed if one direction passes it
> 
> not sure I understand.. why is entire connection "passed" ?can't it be passed in ingress, and allowed in egress?

yes it is, by otherwise, I meant that if it didn't match the first two points ; and is passed on one direction the it is passed.
and it is not for "entire" conns; it is for a connection ( specific)

for example : 
if egress rules allow TCP80; and ingress rule passes TCP80 
then TCP80 is not allowed by both (does not match first point)
it is not denied on one (not matching second point)
and it matches the 3rd point - so it will be passed to be determined by "lower" layers policies

adisos:
> > > otherwise, a connection is passed if one direction passes it
> > 
> > 
> > not sure I understand.. why is entire connection "passed" ?can't it be passed in ingress, and allowed in egress?
> 
> yes it is, by otherwise, I meant that if it didn't match the first two points ; and is passed on one direction the it is passed. and it is not for "entire" conns; it is for a connection ( specific)
> 
> for example : if we have an ANP : if egress rules allow TCP80; and ingress rule passes TCP80 then TCP80 is not allowed by both (does not match first point) it is not denied on one (not matching second point) and it matches the 3rd point - so it will be passed to be determined by "lower" layers policies.
> 
> those "allowed, denied and passed " are related only to the results after looping all ingress and egress rules of ANPs;

by "entire conn" I meant the combined result of both ingress and egress, within a specific connection of course.

>  so it will be passed to be determined by "lower" layers policies.
shouldn't be passed only for ingress flow? and be allowed for egress flow?

shireenf-ibm:
code fixes and changes are done in `eval\check.go` file in last [commit](https://github.com/np-guard/netpol-analyzer/pull/380/commits/9ad15c5fdb086ba1db1fea5462556844eb9af08a) 

thanks to @adisos , I found a bug in the implemented approach;
bug: the allowed connections between two workloads was computed by treating each type of policies as a separated layer; i.e computed the result of combining (intersection of) ingress and egress rules on that layer, then moved to the next layer if needed.

fix: the correct approach is to examine separately :
 - the egress flow throughout all types of policies (considering policies' precedence as previously done)
- the ingress flow throughout all types of policies (considering their precedence)
and finally the allowed connection between the pods is the intersection between allowed conns on egress and allowed conns on ingress.

running the new tests with this fix : 
- all actual results are same as the results that I got from running those tests with policy-assistant.
- some results from old tests were changed, I've confirmed that the new results are the correct ones.

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,129 @@
...
+// PolicyConnections : stores connections between two peers which is collected mainly from admin-network-policies and adjusted with
+// connections from network-policies/ baseline-admin-network-policies or system-default allowed connections as following:
+// - PASS connections will be enforced by allowed connections from network-policies, or if not captured by network-policies,
+// by connections from baseline-admin-network-policies or system-default connections.
+// - traffic that has no match in admin-network-policies will be determined also by network-policies or if not captured by network-policies,
+// by baseline-admin-network-policies or system default.
+type PolicyConnections struct {
+	// AllowedConns allowed connections-set between two peers
+	AllowedConns *common.ConnectionSet
+	// PassConns connections between two peers that was passed by admin-network-policy to policies with lower priority
```
adisos, line 26: documentation on `pass` from `apis/v1alpha1/adminnetworkpolicy_types.go`:
```
Pass: instructs the selected traffic to skip any remaining ANP rules, and
 then pass execution to any NetworkPolicies that select the pod.
 If the pod is not selected by any NetworkPolicies then execution
 is passed to any BaselineAdminNetworkPolicies that select the pod.
```
Thus, using here the term " policies with lower priority" is confusing, as the term "priority" is relevant for ANP as a separate field.
I would change comment, to "delegate decision to next layer of policies, NetworkPolicies or BaselineAdminNetworkPolicies    resources"

adisos: Is there a test that covers a case with "pass" rule after which there are other rules with different action, that demonstrates that the decision is delegated to next layer of policies? (thus making the next ANP rule after "pass" redundant)?

adisos (commented):
few initial comments

pkg/netpol/eval/check.go
```diff
@@ -143,30 +144,56 @@ func (pe *PolicyEngine) AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPe
...
-	var res *common.ConnectionSet
 	var err error
 	// cases where any connection is always allowed
 	if isPodToItself(srcK8sPeer, dstK8sPeer) || isPeerNodeIP(srcK8sPeer, dstK8sPeer) || isPeerNodeIP(dstK8sPeer, srcK8sPeer) {
 		return common.MakeConnectionSet(true), nil
 	}
-	// egress
-	res, err = pe.allAllowedXgressConnections(srcK8sPeer, dstK8sPeer, false)
+
+	// first get conns between src and dst from AdminNetworkPolicies, unless one peer is IP, skip, since ANPs are a cluster level resources
```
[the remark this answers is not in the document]
shireenf-ibm: this comment was deleted in an old commit (it is not in the last version of check.go)

pkg/netpol/eval/check.go
```diff
@@ -397,45 +352,89 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
 
+// analyzing network-policies for conns between peers (object kind == NetworkPolicy):
+
 // getAllAllowedXgressConnsFromNetpols returns if connections from src to dst are captured by network policies on given direction,
 // if yes, returns also the set of allowed connections from src to dst on given direction(ingress/egress), by network policies rules.
 // also checks and updates if a src is exposed to all namespaces on egress or dst is exposed to all namespaces cluster on ingress
-func (pe *PolicyEngine) getAllAllowedXgressConnsFromNetpols(src, dst k8s.Peer, isIngress bool) (allowedConns *common.ConnectionSet,
+// note that network-policies connections represent only allowed conns.
+// note that: if there are policies selecting src (on egress) or dst (on ingress), then the xgress connection is captured;
+// since NetworkPolicy rules implicitly deny unmentioned connections.
```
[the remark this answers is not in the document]
shireenf-ibm: this comment means that :
a networkpolicy that selects the relevant pod is capturing all the connections (explicitly allows the ports in the rules and implicitly denies all other connections);
i.e. a connection is captured if a netpol selecting one of the peers; even if the rules don't select the other peer explicitly.

(in admin-netpols, there is no implicit deny)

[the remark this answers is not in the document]
shireenf-ibm: https://network-policy-api.sigs.k8s.io/api-overview/#adminnetworkpolicy-actions

https://network-policy-api.sigs.k8s.io/api-overview/#general-notes

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,129 @@
...
+	// (network-policies/ baseline-admin-network-policies)
+	PassConns *common.ConnectionSet
+	// DeniedConns denied connections between two peers
+	DeniedConns *common.ConnectionSet
+}
+
+// InitEmptyPolicyConnections - returns a new PolicyConnections object with empty connection-sets
+func InitEmptyPolicyConnections() *PolicyConnections {
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/policy_connections.go
shireenf-ibm: > documentation on `pass` from `apis/v1alpha1/adminnetworkpolicy_types.go`:
> 
> ```
> Pass: instructs the selected traffic to skip any remaining ANP rules, and
>  then pass execution to any NetworkPolicies that select the pod.
>  If the pod is not selected by any NetworkPolicies then execution
>  is passed to any BaselineAdminNetworkPolicies that select the pod.
> ```
> 
> Thus, using here the term " policies with lower priority" is confusing, as the term "priority" is relevant for ANP as a separate field. I would change comment, to "delegate decision to next layer of policies, NetworkPolicies or BaselineAdminNetworkPolicies resources"

done

shireenf-ibm: > Is there a test that covers a case with "pass" rule after which there are other rules with different action, that demonstrates that the decision is delegated to next layer of policies? (thus making the next ANP rule after "pass" redundant)?

yes 
example:
`tests\anp_test_4_with_priority_chang_pass_to_banp` : has 2 ANPs with same connections;
one with `pass` the second with `deny` (the `pass` has precedence) so we see that the BANP determined the conn to be allowed.

and :
`tests\anp_test_4` is the same test; but the ANP with `deny` has the precedence; 

i have added a `diff` tests with both so we really see that the connections are denied in one and allowed in the other one.
(both are have connlist outputs too)

pkg/netpol/eval/check.go
```diff
@@ -436,3 +351,277 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
+	// from the admin netpols capturing the src (if !isIngress)/ capturing the dst (if isIngress true).
+	// connections are computed considering ANPs priorities (rules of an ANP with lower priority take precedence on other ANPs rules)
+	// and rules ordering in single ANP (coming first takes precedence).
+	for _, anp := range pe.sortedAdminNetpols {
+		singleANPConns := k8s.InitEmptyPolicyConnections()
+		// collect the allowed, pass, and denied connectivity from the relevant rules into policiesConns
+		if !isIngress { // egress
+			selectsSrc, err := anp.Selects(src, false)
+			if err != nil {
+				return nil, false, err
```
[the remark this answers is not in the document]
shireenf-ibm: not yet; tests with errors are planned to be in a separate PR (see tasks list [here](https://github.com/np-guard/netpol-analyzer/pull/380#issuecomment-2214896915))

pkg/netpol/eval/internal/k8s/policy_connections.go
adisos: thanks, will take a look.
Is there an example with both rules (pass and another one which is not pass ) in the same policy?

shireenf-ibm: yes (at least one), 
example:
`anp_test_7` : there is a "deny" rule from slytherin and a `pass` rule from slytherin : the result is deny since the "deny" rule is coming first
but in `anp_test_7_swapping_rules` : those rules were swapped, so the `pass` rule is coming first and the result is `allow` (passed to system-default)
- also here there is a `diff` test showing the results

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,348 @@
...
+//
+// a useful way to skip the errors is to define an interface with some Getter funcs to be implemented on the "inheriting types"
+// of the parameters.
+// but in this case : since our parameters are not of local types, we can not define new methods (like getters) on them;
+// not even with using aliases since then we'll need to copy values in calling funcs into the aliases and
+// this is not more efficient than current solutions.
+//
+// @todo: if GO is upgraded to a release that does not has this restriction on types with same fields, replace following two "duplicated"
+// funcs with one func that uses generics type
+
```
adisos, line 93: move this to an issue, and remove comment from here

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,348 @@
...
+// egressRuleSelectsPeer checks if the given []AdminNetworkPolicyEgressPeer rule selects the given peer
+// currently supposing a single egressPeer rule may contain only Namespaces/ Pods Fields,
+// @todo support also egress rule peer with Networks field
+// @todo if egress rule peer contains Nodes field, raise a warning that we don't support it
```
adisos, line 97: why cannot add a warning in this PR?

pkg/netpol/eval/check.go
```diff
@@ -436,3 +351,277 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
+		if err != nil {
+			return nil, err
+		}
+		// if the banp selects the dst on ingress, get ingress conns
+		if selectsDst {
+			res, err = pe.baselineAdminNetpol.GetIngressPolicyConns(src, dst)
+			if err != nil {
+				return nil, err
+			}
+		} else { // egress (!isIngress)
```
adisos, line 609: this `else` is not in the level of `if isIngress { // ingress`

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: currently there is no logger in this pkg (k8s); prefer to add it in a new PR (also in the tasks list above)

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: done, new issue #425

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -0,0 +1,348 @@
...
+		}
+	}
+	return res, nil
+}
+
+const (
+	// according to this: https://network-policy-api.sigs.k8s.io/api-overview/#adminnetworkpolicy-priorities
+	// The Priority field in the ANP spec is defined as an integer value within the range 0 to 1000
+	minPriority = 0
+	maxPriority = 1000
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/policy_connections.go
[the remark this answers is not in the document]
shireenf-ibm: yes if the "passed" connection was not restricted nether by regular netpol, neither by BANP : then it is not covered and will be "converted" to be allowed as a system default allow (in `check.go`).

example : 
we have on the resorces:
ANP passes `TCP80` from `x->y` (policy_conns: allow: empty, deny: empty, pass: TCP80)
there is no NP capturing this connection   --- (delegate to banp)
there is a BANP that capturing egress  from x to y but only on UDP52 (allow) (policy-conns of banp: allow: UDP52, deny empty, pass: empty)

returned result from this fucn : (allow: UDP 52, deny : empty, pass: TCP80)
in the calling func in `check.go lines 423-425` we see that the pass is not empty and add it to the allowed, since what is not restricted is allowed by default

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,129 @@
...
+	pc.AllowedConns.Union(newAdminPolicyConns.AllowedConns)
+	pc.PassConns.Union(newAdminPolicyConns.PassConns)
+}
+
+// CollectConnsFromLowerPrecedencePolicyType updates current PolicyConnections object with connections from a
+// policy with lower priority than ANP. (e.g. network-policy or baseline-admin-network-policy or instead system-default connection)
+// allowed and denied connections of current PolicyConnections object (admin-network-policy) are non-overridden.
+// but pass connections in current PolicyConnections object will be determined by the input PolicyConnections parameter.
+// note that: passConns in otherConns will always be empty. (np and banp don't contain pass connections)
+func (pc *PolicyConnections) CollectConnsFromLowerPrecedencePolicyType(otherConns *PolicyConnections) {
```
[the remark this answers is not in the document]
shireenf-ibm: > * are all `pass` connections expected to be overridden in this function?
yes

1. if the input connection is coming from a netpol then yes; since if it is not allowed by the netpol it is definitely (implicitly) denied.

2. if the input connection is coming from a `banp` : then it may be not mentioned in the banp and then it will be treated as default allow (when returning to the calling func - see [comment](https://github.com/np-guard/netpol-analyzer/pull/380#discussion_r1812039290) 
  
> 
>     * do denied connections include implicit deny from NetworkPolicy resources ?

no and there is no need for that.
if the input connection is coming from an NetworkPolicy.
(current is always from an ANP) 
the flow in this func (as commented and implemented):
1. subtract ANP.allow and ANP.Deny from the NP.allow; so:
what is allowed by the ANP will be kept for sure
what is denied by the ANP will be kept for sure
2. the "remaining" allowed by NP will be union with the allowed by ANP 
here if there are pass by ANP which are allowed by the NP the will be automatically in and allowed 
what is passed by the ANP and not allowed by the NP will not be contained in the allowed 
and in the calling func in `check.go 391-392` the result will only consider the allowed and denied (so every thing is covered) 
--------
please note that this func may not be called twice for same connection;
if the Networkpolicy captures the connection there will be no need to delegate to the BANP .

and all this is explained also in the calling func in `check.go : allAllowedXgressConnections` (see lines 382 - 393 for np capturing the conn with anp) 


>     * what is assumed about input otherConns?
as mentioned in the func docs : otherConns is from a policy in a lower layer than the ANP (either from  NP or from  BANP)
and that passConns in otherConns will always be empty. (np and banp don't contain pass connections)
and more is explained in the calling function in `check.go`

pkg/netpol/eval/check.go
adisos: better add a test covering this if-else, that would catch this kind of bug.. also try to check tests coverage

shireenf-ibm: actually it is covered by the "tests\anp_banp_core_test"

shireenf-ibm: when fixing the bug relevant diff and connlist results of this test were changed

pkg/netpol/eval/resources.go
```diff
@@ -36,6 +38,10 @@ type (
 		netpolsMap                      map[string]map[string]*k8s.NetworkPolicy // map from namespace to map from netpol name to its object
 		podOwnersToRepresentativePodMap map[string]map[string]*k8s.Pod           // map from namespace to map from pods' ownerReference name
 		// to its representative pod object
+		adminNetpolsMap    map[string]bool           // set of input admin-network-policies names to ensure uniqueness by name
+		sortedAdminNetpols []*k8s.AdminNetworkPolicy // sorted by priority list of admin-network-policies;
+		// sorting ANPs occurs after all input k8s objects are inserted to the policy-engine
+		baselineAdminNetpol    *k8s.BaselineAdminNetworkPolicy // pointer to BaselineAdminNetworkPolicy which is a cluster singleton object
```
adisos, line 44: where is it documented in the spec that `BaselineAdminNetworkPolicy` is a singleton?

the documentation of `BaselineAdminNetworkPolicyIngressRule` states that `A total of 100 Ingress rules will be allowed in each BANP instance.` , which may hint there could be more than one instance?

pkg/netpol/eval/internal/k8s/policy_connections.go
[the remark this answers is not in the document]
adisos: > if the input connection is coming from a netpol then yes; since if it is not allowed by the netpol it is definitely (implicitly) denied.

not sure I follow how it is overridden, in case of implicit deny.. the replacement of the `pass` is only for intersection with explicit allow/deny intersections.

pkg/netpol/eval/resources.go
shireenf-ibm: 1. [go resources](https://pkg.go.dev/sigs.k8s.io/network-policy-api@v0.1.5/apis/v1alpha1#BaselineAdminNetworkPolicy)
see in the description:
> message="Only one baseline admin network policy with metadata.name=\"default\" can be created in the cluster" BaselineAdminNetworkPolicy is a cluster level resource that is part of the AdminNetworkPolicy API. 

2. from [openshift-resource](https://www.redhat.com/en/blog/using-adminnetworkpolicy-api-to-secure-openshift-cluster-networking)

> The BaselineAdminNetworkPolicy resource is a cluster singleton object and acts mostly as the default catch-all guardrail policy in case a passed traffic does not match any NetworkPolicies in the cluster. You must use default as the name when creating a BaselineAdminNetworkPolicy object.

pkg/netpol/eval/check.go
```diff
@@ -436,3 +351,277 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
+	// - anpCaptured is true iff there is at least one rule in the input ANPs that captures both src and dst;
+	// because anp rules are read as is and don't contain any implicit isolation effects for the Pods selected by the AdminNetworkPolicy.
+	anpConns, anpCaptured, err := pe.getAllAllowedXgressConnectionsFromANPs(src, dst, isIngress)
+	if err != nil {
+		return nil, err
+	}
+	// second get the allowed xgress conns between the src and dst from the netpols
+	// note that :
+	// - npConns contains only allowed connections
+	// - npCaptured is true iff there are policies selecting either src or dst - since network-policies' rules contain
```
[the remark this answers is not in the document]
shireenf-ibm: fixed

pkg/netpol/eval/check.go
```diff
@@ -436,3 +351,277 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
+// allowed conns are computed by walking through all the available resources of k8s network policy api:
+// admin-network-policies, network-policies and baseline-admin-network-policies;
+// considering the precedence of each policy
+func (pe *PolicyEngine) allAllowedXgressConnections(src, dst k8s.Peer, isIngress bool) (allowedConns *common.ConnectionSet, err error) {
+	// first get allowed xgress conn between the src and dst from the ANPs
+	// note that:
+	// - anpConns may contain allowed, denied or/and passed connections
+	// - anpCaptured is true iff there is at least one rule in the input ANPs that captures both src and dst;
+	// because anp rules are read as is and don't contain any implicit isolation effects for the Pods selected by the AdminNetworkPolicy.
+	anpConns, anpCaptured, err := pe.getAllAllowedXgressConnectionsFromANPs(src, dst, isIngress)
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/internal/testutils/parsed_resources_tests.go
```diff
@@ -0,0 +1,1946 @@
...
+	k8sObj := parser.K8sObject{}
+	k8sObj.Kind = "BaselineAdminNetworkPolicy"
+	k8sObj.BaselineAdminNetworkPolicy = banp
+	return k8sObj
+}
+
+//////////////////////////////////// The following tests are taken from /////////////////////////////////////
+// https://github.com/kubernetes-sigs/network-policy-api/blob/main/cmd/policy-assistant/test/integration/integration_test.go
+
+var (
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/policy_connections.go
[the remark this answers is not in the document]
shireenf-ibm: simplified func by splitting into two (one for np case ; other for banp case)

pkg/netpol/eval/check.go
```diff
@@ -436,3 +351,271 @@ func GetPeerExposedTCPConnections(peer Peer) *common.ConnectionSet {
...
+	// - npConns contains only allowed connections
+	// - npCaptured is true iff there are policies selecting either src or dst based on the checked direction (ingress/ egress)
+	npConns, npCaptured, err := pe.getAllAllowedXgressConnsFromNetpols(src, dst, isIngress)
+	if err != nil {
+		return nil, err
+	}
+
+	// compute the allowed connections on the given direction considering the which policies captured the xgress connection
+	// and precedence of each policy type:
+	switch npCaptured {
```
adisos, line 385: the `switch` can be rewritten for better clarity:
```
switch{
  case npCaptured && !anpCaptured : 
   ...
  case npCaptured && anpCaptured : 
   ...
  case !npCaptured:
   ...
}
```

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,151 @@
...
+	// also, banpConns.DeniedConns currently contains:
+	// 1. traffic that was passed by ANPs (if there are such conns)
+	// 2. or traffic that had no match in ANPs
+	// so we can update current denied conns with banpConns.DeniedConns
+	pc.DeniedConns.Union(banpConns.DeniedConns)
+	// now Pass conns were handled automatically; pc.PassConns is not relevant anymore.
+	// Pass Conns which are not captured by BANP, will be handled now with all other conns
+	// all conns that are not determined by the ANP and BANP are allowed by default
+	allowedByDefault := common.MakeConnectionSet(true)
+	allowedByDefault.Subtract(pc.DeniedConns)
```
adisos, line 135: I think it's a bit confusing and not accurate to name this var `allowedByDefault`, because not every connection that is not caught as "denied" by ANP, is considered as "allowed by default", it could also be caught as "allowed" by ANP.

shireenf-ibm: renamed to `nonCapturedConns` (not captured by policies)

pkg/netpol/eval/internal/k8s/policy_connections.go
```diff
@@ -0,0 +1,151 @@
...
+func (pc *PolicyConnections) IsEmpty() bool {
+	return pc.AllowedConns.IsEmpty() && pc.DeniedConns.IsEmpty() && pc.PassConns.IsEmpty()
+}
+
+// DeterminesAllConns : returns true if the allowed and denied connections of the current PolicyConnections object
+// selects all the connections
+func (pc *PolicyConnections) DeterminesAllConns() bool {
+	selectedConns := pc.AllowedConns.Copy()
+	selectedConns.Union(pc.DeniedConns)
+	return selectedConns.AllConnections()
```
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/check.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/policy_connections.go
adisos: but it may include connections captured by ANPs..

shireenf-ibm: true (those conns are already in `pc.AllowedConns`)
this variable purpose is to add the conns which were not captured at all (so they are allowed by default).
do you have another suggestion?

shireenf-ibm: i can subtract the pc.AllowedConns from it (but they are union-ed at the end, so the subtract is redundant -  meaningless )

adisos: few comments/questions:

- as for the name, it should indicate what it contains, which is all connections but (anp/banp)-denied, so a short name could be `allButDenied` with a comment explaining this.
- about the function itself: isn't it sufficient to just compute the following in the function body instead:
```
banpConns.DeniedConns.Subtract(pc.AllowedConns)
pc.DeniedConns.Union(banpConns.DeniedConns)
pc.AllowedConns = common.MakeConnectionSet(true).Subtract(pc.DeniedConns)
```

shireenf-ibm: you are right those lines actually does the same, will update it

shireenf-ibm: done

pkg/internal/netpolerrors/netpol_errors.go
```diff
@@ -126,3 +141,25 @@ const colonSep = ": "
...
+
+// SamePriorityErr returns the error message if a priority appears more than once in different admin-network-policies
+func SamePriorityErr(name1, name2 string) string {
+	return "Admin Network Policies: " + name1 + " and " + name2 + " have same priority;" +
+		"Two policies are considered to be conflicting if they are assigned the same priority."
+}
+
+// PriorityValueErr returns error message of invalid priority value in an admin-network-policy
+func PriorityValueErr(name string, priority int32) string {
+	return fmt.Sprintf("Invalid Priority Value: %d in Admin Network Policy: %q; Priority value must be between 0-1000", priority, name)
```
adisos, line 153: use const values for valid priority values instead of hard-coded ones

shireenf-ibm: done

adisos: why isn't pkg netpolerrors under `pkg/netpol/internal` ?
I think it may be better to define these consts under pkg named `common` and not under pkg `netpolerrors`

shireenf-ibm: `netpolerrors` is used also in `cli` and `manifests` (can not be moved to `netpol\internal`)
i agree `netpolerrors` it is not the best pkg to put those constants. (but `netpolerrors` is the only seen to both now, don't think worth to add new pkg because of two contants.)

shireenf-ibm: done, new issue #446

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1139,4 +1140,229 @@ var goodPathTests = []struct {
...
+	},
+	{
+		testDirName:   "anp_test3_pass_traffic",
+		outputFormats: []string{output.TextFormat},
+	},
+	{
+		// Should Deny traffic from slytherin to gryffindor and
+		// Deny traffic to slytherin from gryffindor respecting ANP with priority 50, ignoring ANP with priority 60
+		testDirName:   "anp_test_4",
+		outputFormats: ValidFormats,
```
adisos, line 1160: please use consistent test names, rename `anp_test1_deny...` to be `anp_test_1_deny...` (also for anp_test2 and anp_test3)

pkg/netpol/connlist/connlist_test.go
```diff
@@ -1139,4 +1140,229 @@ var goodPathTests = []struct {
 		testDirName:   "netpol_named_port_test_2",
 		outputFormats: []string{output.DefaultFormat},
 	},
+	// tests with adminNetworkPolicy
+	{
+		testDirName:   "anp_test1_deny_traffic_at_cluster_level",
+		outputFormats: []string{output.TextFormat},
+	},
+	{
+		testDirName:   "anp_test2_allow_traffic_at_cluster_level",
```
adisos, line 1149: for test `tests/anp_test2_allow_traffic_at_cluster_level`, the result connectivity will be the same even if the policy is deleted, so the effect of the policy is not well demonstrated.
better change the test so that the effect of the tested policy is actual.
please also check this for all other tests.

shireenf-ibm: okay, I remember those 3 were the first tests I've added ; were taken from [here](https://network-policy-api.sigs.k8s.io/reference/examples/).
will add `NetworkPolicy` that blocks some of the traffic and see that it is allowed because of the ANP.

will check for other tests

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: done

pkg/netpol/connlist/connlist_test.go
shireenf-ibm: done

adisos: not sure I understand `anp_test_3` - the output with and without the anp is the same

shireenf-ibm: true, I jsut wanted to show that if the anp passes a connection but this connection is not allowed by networkpolicy then it is not allowed.
I could add A BANP that denies TCP 8080 ; but we have such example

shireenf-ibm: done

adisos: test  `tests/anp_test_6_swapping_rules`, rule  `"deny-from-slytherin-at-port-5353"`  - better add another port for this rule, so the output shows the effect of both "pass" and "deny" rules instead of juts "pass". (so that only part of the deny rule is overridden and not entirely).

shireenf-ibm: done

tests/anp_test_ingress_egress_intersection/anp.yaml
```diff
@@ -0,0 +1,25 @@
...
+            port: 5353
+    - action: Deny # deny others
+      from:
+      - namespaces: {}
+  egress:
+    - action: Allow
+      to:
+      - namespaces:
+            matchLabels:
+              kubernetes.io/metadata.name: foo-ns-2
```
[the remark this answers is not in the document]
shireenf-ibm: yes, and it was intended;
 to see that even though the "egress" rule allows all conns to the "foo-ns-2"; only "UDP 3535" is allowed because of the ingress rule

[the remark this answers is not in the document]
adisos: ok, maybe this test can be improved with an "egress" for "udp 5000-6000" , and then the intersection from ingress,egress is only "udp 5353".
can also add another port on the ingress rule, so that it's a non trivial intersection.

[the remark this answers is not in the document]
shireenf-ibm: done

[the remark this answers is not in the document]
shireenf-ibm: TCP 2000 is allowed also on egress,
ANP rules are read as is , it is not mentioned in the rule but not restricted by any other egress rule/policy. so it is allowed by default on egress and since allowed also by ingress it is allowed.
(the policy is still in a mode that the "egress" section does not affect the result.)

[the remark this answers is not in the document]
shireenf-ibm: i can add a "deny all" rule to egress; then the intersection will not include TCP 2000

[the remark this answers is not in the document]
shireenf-ibm: done

PR #395, opened 2024-08-11
Adding complement string representation for conns

shireenf-ibm:
#380 (base PR)
issue #441 

* in this PR, adding a complement for string representation of a connection-set.
for example: writing `All but: UDP 5353` instead of `"SCTP 1-65535,TCP 1-65535,UDP 1-5352,5354-65535"`

* currently: the decision if to use the connectionSet string as is or its complement string is choosing the shorter string between the two.


adisos:
@shireenf-ibm  Le'ts close this PR for now, and decide later if/when this enhancement should be added.

shireenf-ibm:
to be decided later when to add this enhancement

Issue #405, opened 2024-09-03
fix a bug in getting netpol connections from  rule  with named port and protocol


shireenf-ibm:
currently, if a netpol rule contains a named-port + a protocol name; 
the named-port is converted to its port number in the `dst` pod without checking if both rule's protocol and pod's protocol (for the relevant port) are equal too - which might result a wrong output

example with the error :
suppose having  `pod-a` with following named-port : 
```
- name: newport
   containerPort: 8956
   protocol: SCTP
```

and a netpol capturing this `pod-a` with an ingress rule (from any other pod (e.g. `new-pod`)) with following ports:
```
    ports:
    - port: newport
      protocol: UDP
```

in the result we would see a `wrong` connection to the pod :
`helloworld/new-pod[Deployment] => helloworld/pod-a[Deployment] : UDP 8956`

while actually there is no match between the rule's port and the pod's named-port, and we shouldn't see a connection.

example files:
[netpol.yaml.txt](https://github.com/user-attachments/files/16846040/netpol.yaml.txt)
[pods.yaml.txt](https://github.com/user-attachments/files/16846043/pods.yaml.txt)

PR #412, opened 2024-09-18
bug fix and tests

shireenf-ibm:
#405

bug(s) description:
1. the analyzer ignored netpol rules with `named ports` that didn't have a match in the dst pod's configuration 
(for example: if an ingress netpol captures `pod-a`  with ingress rule that contains a named-port `port-a`  (TCP protocol) and the pod's specification does not has a containerPort/ Port with this name -- this rule was ignored, has no mention in the output)
   - **fix:** now we will see an  ingress conn to `pod-a` on: `TCP port-a` in the output
   
 2. when converting named-port to number, the analyzer didn't compare netpol rule's protocol (which is accompanied with the port) to the protocol of the pod's container port which has same port-name.
 for example: if a pod has containerPort with `name: newport` and `protocol: UDP port:90` 
 and a netpol capturing that pod with ingress rule: `port: newport protocol:TCP`  
 we saw on the output an ingress connection `TCP 90` which is not true since this is not the same port.
     - **fix:** : the port-name and protocol are considered when converting a named port; if protocols don't match we will see the port-name in the output (in the example: conn is `TCP newport`)
 
3. `Peer2PeerConnection`  (`Connection`) assumed the ports will always contain a numbered interval only
     - **fix:** : `PortRange` interface contains also a `NamedPort` option; (the connection on a protocol may either be numbered or named; depends if we succeeded to convert it by from the pod's configuration or not)
     
     
tests with examples of the above (with matched and unmatched named-ports) were added


adisos:
> #405
> 
> bug(s) description:
> 
>     1. the analyzer ignored netpol rules with `named ports` that didn't have a match in the dst pod's configuration
>        (for example: if an ingress netpol captures `pod-a`  with ingress rule that contains a named-port `port-a`  (TCP protocol) and the pod's specification does not has a containerPort/ Port with this name -- this rule was ignored, has no mention in the output)
>        
>        * **fix:** now we will see an  ingress conn to `pod-a` on: `TCP port-a` in the output
> 
>     2. when converting named-port to number, the analyzer didn't compare netpol rule's protocol (which is accompanied with the port) to the protocol of the pod's container port which has same port-name.
>        for example: if a pod has containerPort with `name: newport` and `protocol: UDP port:90`
>        and a netpol capturing that pod with ingress rule: `port: newport protocol:TCP`
>        we saw on the output an ingress connection `TCP 90` which is not true since this is not the same port.
>        
>        * **fix:** : the port-name and protocol are considered when converting a named port; if protocols don't match we will see the port-name in the output (in the example: conn is `TCP newport`)
> 
>     3. `Peer2PeerConnection`  (`Connection`) assumed the ports will always contain a numbered interval only
>        
>        * **fix:** : `PortRange` interface contains also a `NamedPort` option; (the connection on a protocol may either be numbered or named; depends if we succeeded to convert it by from the pod's configuration or not)
> 
> 
> tests with examples of the above (with matched and unmatched named-ports) were added


I agree with `2` , but not sure about  `1` , whether it is an issue. (Does `3` follow from `1` ?)
As for `1` -- if a concrete Pod has no matching named port on the yaml config, can it accept traffic on such named port that the policy specifies?
(For exposure analysis it makes sense to specify named ports, since we do not have the real pod configuration).

pkg/netpol/connlist/connlist.go
```diff
@@ -349,8 +350,12 @@ func GetConnectionSetFromP2PConnection(c Peer2PeerConnection) *common.Connection
 	protocolsToPortSetMap := make(map[v1.Protocol]*common.PortSet, len(c.ProtocolsAndPorts()))
 	for protocol, portRageArr := range c.ProtocolsAndPorts() {
 		protocolsToPortSetMap[protocol] = common.MakePortSet(false)
-		for _, p := range portRageArr {
-			protocolsToPortSetMap[protocol].AddPortRange(p.Start(), p.End())
+		for _, p := range portRageArr { // each single port range may contain either named port or an interval of start and end numbers
```
adisos, line 353: rename `portRageArr` to `portRangeArr`

pkg/netpol/connlist/internal/ingressanalyzer/ingress_analyzer.go
```diff
@@ -360,11 +360,12 @@ func (ia *IngressAnalyzer) getIngressPeerConnection(peer eval.Peer, actualServic
 	for _, peerPortToFind := range peerPortsToFind {
 		portNum := peerPortToFind.IntValue()
 		if peerPortToFind.StrVal != "" { // if the port we are searching for is namedPort
-			portInt, err := ia.pe.ConvertPeerNamedPort(peerPortToFind.StrVal, peer)
+			protocol, portInt, err := ia.pe.ConvertPeerNamedPort(peerPortToFind.StrVal, peer)
 			if err != nil {
 				return nil, err
 			}
-			if portInt < 0 { // no matching port for the given named port
+			// only TCP ports are acceptable
```
adisos, line 367:
```suggestion
			// only TCP ports are acceptable for Ingress resource
```

pkg/netpol/connlist/connlist.go

shireenf-ibm:
> > #405
> > bug(s) description:
> > ```
> > 1. the analyzer ignored netpol rules with `named ports` that didn't have a match in the dst pod's configuration
> >    (for example: if an ingress netpol captures `pod-a`  with ingress rule that contains a named-port `port-a`  (TCP protocol) and the pod's specification does not has a containerPort/ Port with this name -- this rule was ignored, has no mention in the output)
> >    
> >    * **fix:** now we will see an  ingress conn to `pod-a` on: `TCP port-a` in the output
> > 
> > 2. when converting named-port to number, the analyzer didn't compare netpol rule's protocol (which is accompanied with the port) to the protocol of the pod's container port which has same port-name.
> >    for example: if a pod has containerPort with `name: newport` and `protocol: UDP port:90`
> >    and a netpol capturing that pod with ingress rule: `port: newport protocol:TCP`
> >    we saw on the output an ingress connection `TCP 90` which is not true since this is not the same port.
> >    
> >    * **fix:** : the port-name and protocol are considered when converting a named port; if protocols don't match we will see the port-name in the output (in the example: conn is `TCP newport`)
> > 
> > 3. `Peer2PeerConnection`  (`Connection`) assumed the ports will always contain a numbered interval only
> >    
> >    * **fix:** : `PortRange` interface contains also a `NamedPort` option; (the connection on a protocol may either be numbered or named; depends if we succeeded to convert it by from the pod's configuration or not)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > tests with examples of the above (with matched and unmatched named-ports) were added
> 
> I agree with `2` , but not sure about `1` , whether it is an issue. (Does `3` follow from `1` ?) As for `1` -- if a concrete Pod has no matching named port on the yaml config, can it accept traffic on such named port that the policy specifies? (For exposure analysis it makes sense to specify named ports in the output, since we do not have the real pod configuration).

reverted 1 & 3

shireenf-ibm:
* deleted tests that does not have fixed results (comparing to `main`).

remaining tests results fix :
- test: `netpol_named_port_test`

output with bug included following line: (`UDP 8956` is a wrong convert of pod's namedPort: `- name: newport
          containerPort: 8956
          protocol: SCTP`
`helloworld/new-pod[Deployment] => helloworld/pod-a[Deployment] : UDP 8956`
after fixing the bug this line will not be included since`pod-a` configuration does not contain a `newport: 8956` with `UDP`

- netpol_named_port_test_2: 

output with bug would have following line:
`helloworld/pod-a[Deployment] => helloworld/new-pod[Deployment] : SCTP 8956,9090,**UDP 9090**`
after fixing bug the correct connection is:
`helloworld/pod-a[Deployment] => helloworld/new-pod[Deployment] : SCTP 8956,9090`
(since `new-pod` does not have a port named: `newport: 9090` with protocol UDP`)

tests/netpol_named_port_test_2/netpol.yaml
adisos, line 1: please add comment near each named port in the tests, if it is expected to be matched by network policy rule with named port or not.

also, do we need both tests? what is the difference between what they cover?

shireenf-ibm: added comments to netpols files + connlist_test file.
although the tests cover same "issue"
(both tests contain a named-port with protocol that does not match the protocol in the pod's configuration (with same named-port)
and one test's netpol includes also a named-port that has no matching port name in the pod.)

the output of the tests differs, 
- in one test since there is no matching port - we don't see any connection between the 2 pods in the connlist
- in the second tests, we see connection with the two ports that were successfully converted

adisos (approved):

PR #423, opened 2024-10-20
Admin Network Policy: eval command support 

shireenf-ibm:
#380 adding support to eval command
issue #443 

* in first [commit](https://github.com/np-guard/netpol-analyzer/pull/423/commits/5e900229d54e7b756f657e7def0de61d7dd148b5) :
1. updated funcs related to eval command under `eval` pkg (for my convenience, i have separated the funcs of `check.go` file into 2 files, the new file is `check_eval.go`; all the related funcs to eval command (checking if specific conn is allowed) are there
2.  successfully running unit-tests from parsed resources in `eval_test.go`

* in second [commit](https://github.com/np-guard/netpol-analyzer/pull/423/commits/1e8ccceece3606ff7c3fd0eaed6b5c7899550eed) :
1. added new command-line tests :`eval` command with tests with ANPs (added relevant support for new policies (anps+banp) + for pods that are actually replica of a workload object that is not defined as `Pod` in the yamls (e.g. `StatefulSet`))


pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -234,6 +236,124 @@ func subjectSelectsPeer(anpSubject apisv1a.AdminNetworkPolicySubject, p Peer) (b
...
+	Allow
+	Deny
+)
+
+// determineConnResByAction gets rule action that captured a connection and returns the rule res (allow, pass, deny)
+func determineConnResByAction(action string, isBANPrule bool) (res int, captured bool, err error) {
+	switch action {
+	case string(apisv1a.AdminNetworkPolicyRuleActionPass):
+		if isBANPrule {
+			return -1, false, errors.New(netpolerrors.UnknownRuleActionErr)
```
adisos, line 345: avoid using hard-coded numbers..
better also to create a dedicated type to represent the result instead of `int`, and define by const vars its possible values. (see references for "enums" in Go).

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -234,6 +236,124 @@ func subjectSelectsPeer(anpSubject apisv1a.AdminNetworkPolicySubject, p Peer) (b
...
+	}
+	if !peerSelected {
+		return 0, false, nil
+	}
+	connSelected, err := anpPortContains(rulePorts, protocol, port, dst)
+	if err != nil {
+		return 0, false, err
+	}
+	if !connSelected {
+		return 0, false, nil
```
adisos, line 328: same here.. it is not clear what "-1" or "0" means..

adisos (changes requested):
initial comments

pkg/netpol/eval/check_eval.go
```diff
@@ -0,0 +1,257 @@
...
+// analyzeANPCapturedRes when an admin-network-policy captures a connection , its result may be Allow (final- allowed conn),
+// or Deny (final - denied conn) or Pass (to be determined by netpol/ banp)
+func analyzeANPCapturedRes(anpRes int) (allowedOrDenied, pass bool, err error) {
+	switch anpRes {
+	case k8s.Pass:
+		return false, true, nil
+	case k8s.Allow:
+		return true, false, nil
+	case k8s.Deny:
+		return false, false, nil
```
adisos, line 157: why is the first return arg named `allowedOrDenied`, but for the case of `Deny` it returns `false`?

pkg/netpol/eval/check_eval.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return false, err
+	}
+	pe.cache.addConnectionResult(src, dst, protocol, port, defaultRes)
+	return defaultRes, nil
+}
+
+// allowedXgressConnectionByAdminNetpols returns if the given input connection is allowed on the given ingress/egress direction
+// by k8s admin-network-policies
+func (pe *PolicyEngine) allowedXgressConnectionByAdminNetpols(src, dst k8s.Peer, isIngress bool, protocol, port string) (res, pass bool,
```
adisos, line 108: rename return arg `pass` to `passOrUncaptured` , to avoid confusing "pass" action with "non captured" concept..

pkg/netpol/eval/check_eval.go
```diff
@@ -0,0 +1,257 @@
...
+	pe.cache.addConnectionResult(srcPeer, dstPeer, protocol, port, ingressRes)
+	return ingressRes, nil
+}
+
+// allowedXgressConnection returns if the given input connection is allowed on the given ingress/egress direction
+// by k8s policies api
+func (pe *PolicyEngine) allowedXgressConnection(src, dst k8s.Peer, isIngress bool, protocol, port string) (bool, error) {
+	// first checks if the connection is allowed or denied by the admin-network-policies (return anpRes).
+	// if the connection is passed by the ANPs or not captured by them, then will continue to NPs (pass = true)
+	anpRes, pass, err := pe.allowedXgressConnectionByAdminNetpols(src, dst, isIngress, protocol, port)
```
adisos, line 74: also here - rename var `pass` to be `passOrUncaptured`

pkg/netpol/eval/internal/k8s/netpol.go
```diff
@@ -66,7 +66,7 @@ const (
 )
 
 func getProtocolStr(p *v1.Protocol) string {
-	if p == nil { // If not specified, this field defaults to TCP.
+	if p == nil || string(*p) == "" { // If not specified, this field defaults to TCP.
 		return "TCP"
```
adisos, line 70:
```suggestion
		return string(v1.ProtocolTCP)
```

pkg/netpol/eval/check_eval.go
shireenf-ibm: the name of the variable is  "allowedOrDenied" i.e. the connection is either allowed or denied
in case it is allowed then the result for the given connection is true;
but if it is deny then the result is false

pkg/netpol/eval/internal/k8s/baseline_admin_netpol.go
```diff
@@ -71,3 +71,37 @@ func (banp *BaselineAdminNetworkPolicy) GetIngressPolicyConns(src, dst Peer) (*P
 	}
 	return res, nil
 }
+
+func (banp *BaselineAdminNetworkPolicy) CheckEgressConnAllowed(dst Peer, protocol, port string) (res int, captured bool, err error) {
```
adisos, line 75: missing documentation for these functions

pkg/netpol/eval/internal/k8s/baseline_admin_netpol.go
```diff
@@ -71,3 +71,37 @@ func (banp *BaselineAdminNetworkPolicy) GetIngressPolicyConns(src, dst Peer) (*P
...
+		if !captured {
+			continue
+		}
+		return res, captured, nil
+	}
+	// getting here means the protocol+port was not captured
+	return Allow, false, nil
+}
+
+func (banp *BaselineAdminNetworkPolicy) CheckIngressConnAllowed(src, dst Peer, protocol, port string) (res int, captured bool, err error) {
```
adisos, line 92: here as well

pkg/netpol/eval/internal/k8s/baseline_admin_netpol.go
adisos: why does it need to return `int` and not just `bool` for whether allowed or not? why is `captured` needed?

pkg/netpol/eval/check_eval.go
adisos: ok, please explain in function documentation

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -319,3 +439,39 @@ func (anp *AdminNetworkPolicy) HasValidPriority() bool {
...
+		if err != nil {
+			return 0, false, anp.anpRuleErr(rule.Name, err.Error())
+		}
+		if !captured {
+			continue
+		}
+		return ruleRes, captured, nil
+	}
+	// getting here means the protocol+port was not captured
+	return Pass, false, nil
```
adisos, line 458: confusing const action `Pass` for the case of not captured

pkg/netpol/eval/internal/k8s/adminnetpol.go
```diff
@@ -234,6 +236,124 @@ func subjectSelectsPeer(anpSubject apisv1a.AdminNetworkPolicySubject, p Peer) (b
 	return doesPodsFieldMatchPeer(anpSubject.Pods, p)
 }
 
+// anpPortContains returns if the given AdminNetworkPolicyPort selects the input connection
+//
+//gocyclo:ignore
+func anpPortContains(rulePorts *[]apisv1a.AdminNetworkPolicyPort, protocol, port string, dst Peer) (bool, error) {
```
adisos, line 242: does this function have some code used also in `func (np *NetworkPolicy) ruleConnsContain` that can be captured in helper functions instead?

pkg/netpol/eval/check_eval.go
```diff
@@ -0,0 +1,257 @@
...
+	return defaultRes, nil
+}
+
+// allowedXgressConnectionByAdminNetpols returns if the given input connection is allowed on the given ingress/egress direction
+// by k8s admin-network-policies
+func (pe *PolicyEngine) allowedXgressConnectionByAdminNetpols(src, dst k8s.Peer, isIngress bool, protocol, port string) (res, pass bool,
+	err error) {
+	// iterate sorted by priority admin netpols
+	for _, anp := range pe.sortedAdminNetpols {
+		if isIngress {
```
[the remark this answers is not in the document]
shireenf-ibm: admin netpols are cluster scoped - don't have namespace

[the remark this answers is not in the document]
adisos: maybe divide by the subject namespaces?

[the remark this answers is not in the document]
shireenf-ibm: subject may be by namespaces or by pods (and a pod name may match more than one namespace).
I don't think this would be more efficient than what happens now;
 we'll need to loop the admin-netpols after inserting all namespaces and workloads and then parse their subjects in order to store in a map;

pkg/cli/evaluate.go
```diff
@@ -92,12 +98,33 @@ func updatePolicyEngineObjectsFromDirPath(pe *eval.PolicyEngine, podNames []type
...
+			err = pe.InsertObject(obj.DaemonSet)
+		case parser.StatefulSet:
+			err = pe.InsertObject(obj.StatefulSet)
+		case parser.ReplicationController:
+			err = pe.InsertObject(obj.ReplicationController)
+		case parser.Job:
+			err = pe.InsertObject(obj.Job)
+		case parser.CronJob:
+			err = pe.InsertObject(obj.CronJob)
+			// ns kind
```
adisos, line 118: if `eval` only supports pods, why adding these resources?
If the dir path only contains workload manifests and not pods, how would one infer the pod name for this command?
maybe this should be a separate issue, to extend eval support for workloads, but in this PR can just focus on pods?
the tests automation can add the pods resources and adjust the input, as if those manifests would exist for testing.

shireenf-ibm: not sure understood last sentence;
my tests dir paths contain manifests with StatefulSet workloads (I used replica name as the input of the test (pod name);
The tests failed without adding these cases (my manifests were not inserted to the policy-engine because of the lack of these cases; );
should I revert these changes and create new tests with pods only in the manifests?

adisos: for these tests to work, you can add test automation (for eval tests) that add pods resources as if they existed in the manifests.. the changes here better be removed, until eval actually supports such resources, because in the current eval usage workload resources are not supported, and a user cannot infer the pod name to be used in this command when providing only workload resources.

pkg/netpol/eval/check_eval.go
shireenf-ibm: done

pkg/netpol/eval/check_eval.go
shireenf-ibm: done

pkg/netpol/eval/check_eval.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+		}
+	}
+	// getting here means the connection was not captured by any ANP - pass to netpols
+	return false, true, nil
+}
+
+// analyzeANPCapturedRes when an admin-network-policy captures a connection , its result may be Allow (final- allowed conn),
+// or Deny (final - denied conn) or Pass (to be determined by netpol/ banp)
+func analyzeANPCapturedRes(anpRes int) (allowedOrDenied, pass bool, err error) {
```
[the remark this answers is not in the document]
shireenf-ibm: renamed

pkg/netpol/eval/check_eval.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: changed

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: changed

pkg/netpol/eval/internal/k8s/baseline_admin_netpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/baseline_admin_netpol.go
shireenf-ibm: done

pkg/netpol/eval/internal/k8s/adminnetpol.go
shireenf-ibm: yes, done

pkg/cli/evaluate.go
shireenf-ibm: modified and added tests:
1. `command_test.go` : since `command-line` needs the "dir" as an argument to execute the cmd; 
for those tests (which don't contain pods in the manifests), I've generated a "temp dir" which is a copy of the original dir + generated yamls for pods (for the given src + dst arguments) so the command runs on the "temporary dir" successfully. 
_generating pod yaml for a workload considers also its labels (which may be used in the netpols rules and are essential for computing the result);_ 


2. `eval_test.go` uses the internal `eval` funcs, so it works successfully with the original dirs.
in this file we had tests from parsed resources (that tests the "eval" functionality) ;
 and I have added eval tests from dir paths.

pkg/internal/testutils/generate_pod_yamls.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	// copy orig dir into the temp dir
+	err = copyDir(origDir, testTmpDir, srcName, srcNs, dstName, dstNs)
+	return testTmpDir, err
+}
+
+// copyDir copies files of network-policies from origDir into tempDir
+// and generates into the tempDir : Pod yaml files for given src and dst peers from their workload resources
+// in the origDir
+func copyDir(origDir, tempDir, srcName, srcNs, dstName, dstNs string) error {
```
adisos, line 58: consider renaming `copyDir` to something like `copyAndExtendDir` or `copyDirAndAddPods`

pkg/internal/testutils/generate_pod_yamls.go
```diff
@@ -0,0 +1,204 @@
...
+// the function returns the path of the generated temp dir.
+func GenerateTempDirWithPods(origDir, srcName, srcNs, dstName, dstNs string) (string, error) {
+	// create the TmpDir path if does not exist
+	if _, err := os.Stat(TmpDir); os.IsNotExist(err) {
+		osErr := os.Mkdir(TmpDir, exeMode)
+		if osErr != nil {
+			return "", osErr
+		}
+	}
+	// making temp directory in the temp path
```
adisos, line 45: why is this required? isn't it sufficient to just use `TmpDir` ?

pkg/internal/testutils/generate_pod_yamls.go
```diff
@@ -0,0 +1,204 @@
...
+
+// copyFile copies origFile to tempFile
+func copyFile(origFile, tempFile string) error {
+	contents, err := os.ReadFile(origFile)
+	if err != nil {
+		return err
+	}
+	err = os.WriteFile(tempFile, contents, exeMode)
+	return err
+}
```
adisos, line 169: is there an `os` func that can be used instead?

pkg/internal/testutils/generate_pod_yamls.go
```diff
@@ -0,0 +1,204 @@
...
+	docs := splitYamlDocs(fileContents)
+
+	// we are interested in Metadata of the workload only.
+	// Iterate through objects to find the matching one
+	for _, doc := range docs {
+		var obj WorkloadMetadata
+		if err := yaml.Unmarshal(doc, &obj); err != nil {
+			return nil, err
+		}
+		if obj.Metadata.Name == podName && (obj.Metadata.Namespace == podNs ||
```
adisos, line 135: can use instead the already existing functions from `manifests.GetResourceInfosFromDirPath()` and 
 `parser.ResourceInfoListToK8sObjectsList()` ?

pkg/netpol/eval/check_eval.go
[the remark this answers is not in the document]
adisos: the function name and  return arg names are still confusing.. maybe consider a separate type for capturing all 3 options (allow/deny/pass) of return value ? instead of splitting into two return args, where sometimes their result is not relevant...

pkg/cli/evaluate.go
adisos: ok, please open a separate issue to consider supporting `eval` with input workloads, not just pods, and also add a comment for this test functionality added here, that it can be removed when this `eval` extension is added.

pkg/netpol/eval/eval_test.go
```diff
@@ -1870,3 +1870,105 @@ func TestAllParsedResourcesEvalTests(t *testing.T) {
...
+		},
+		{
+			dir:        "anp_banp_core_test",
+			sourceNs:   "network-policy-conformance-gryffindor",
+			sourcePod:  "harry-potter",
+			destPod:    "cedric-diggory",
+			destNs:     "network-policy-conformance-hufflepuff",
+			port:       "8080",
+			evalResult: true,
+		},
```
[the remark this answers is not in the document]
shireenf-ibm: yes, the names are of the relevant workloads in the manifests.
I mentioned that the input-dir does not contain pods so it will generate a pod matching the workload resource with the src + dst names.
e.g. the generated src pod takes the `metadata` of the workload with the `sourcePod` param name. (i.w. its name, namespace and labels)

pkg/cli/evaluate.go
shireenf-ibm: done, new issue #440 ;  + comments were added in the relevant files

pkg/internal/testutils/generate_pod_yamls.go
```diff
@@ -0,0 +1,204 @@
...
+// a test util for eval command-line; generating a new temporary directory with pod files to be used for testing.
+// eval command supports only test directories with Pod objects.
+// some of our testing directories are containing other Workload types in the manifests; in order to be able to test
+// eval command on those dirs (in unit-testing); this file funcs are used.
+// we copy the test-die into a new temporary dir, and generate into the tempDir :
+// Pod yaml files for given src and dst peers from their workload resources
+
+const (
+	tmpPattern = "temp-*"
+	exeMode    = 0o777
```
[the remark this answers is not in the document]
shireenf-ibm: true, done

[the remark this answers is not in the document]
shireenf-ibm: i put 600 for file access and 700 for creating the dir ( 600 on all failed)

pkg/cli/command_test.go
```diff
@@ -348,14 +348,15 @@ func TestDiffCommandOutput(t *testing.T) {
...
-		evalResult bool
+		dir               string
+		sourcePod         string
+		sourceNs          string
+		destNs            string
+		destPod           string
+		protocol          string
+		port              string
+		evalResult        bool
+		generateManifests bool // indicates if the test dir does not contain pods - to be generated
```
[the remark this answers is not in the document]

pkg/internal/testutils/generate_pod_yamls.go
shireenf-ibm: done

pkg/internal/testutils/generate_pod_yamls.go
shireenf-ibm: I didn't met such func that I can use to copy files in an efficient or "shorter" way

pkg/netpol/eval/eval_test.go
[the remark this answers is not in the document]
adisos: ok, so the test attribute should be `srcWorkload` or `dst`Workload`, and also better add some documentation explaining that for these tests `eval` will be called with pod names owned by these workloads (pods which are added by policy engine ).

pkg/internal/testutils/generate_pod_yamls.go
shireenf-ibm: true, done

pkg/netpol/eval/eval_test.go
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/check_eval.go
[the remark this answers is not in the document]
shireenf-ibm: tried to rename the returned values same as the return values of its calling func and added some documentation; 
is this better?

this func is a helper that returns same values as its calling func `allowedXgressConnectionByAdminNetpols` 
it is used only to avoid duplicate in the calling func.

adisos (commented):
fix a conflict in go.mod

PR #486, opened 2025-01-21
W.I.P Exposure analysis with anp

shireenf-ibm:


pkg/netpol/connlist/exposure_map.go
```diff
@@ -123,14 +123,24 @@ func (ex *exposureMaps) addConnToExposureMap(pe *eval.PolicyEngine, allowedConne
...
+	// 2. if the peer is not exposed to entire-cluster and there is no connection between the peer and representative-peer (no exposure)
+	if equals || (!isExposed && allowedConnSet.IsEmpty()) {
 		return nil // skip
 	}
+
+	// add entry to exposure map if:
+	// 1. the allowed connection between the real peer and representative peer is not empty (and not covered by
+	// entire-cluster conn) - exposure case
+	// 2. if the connection is empty (no conns) and the peer is exposed to entire-cluster - add this empty connection to the exposure map
+	// to tell that this representative peer is excluded from `entire-cluster` connection
```
adisos, line 142: how is it going to be marked as excluded in the output?

pkg/netpol/eval/exposure.go
```diff
@@ -26,13 +26,19 @@ import (
 // - generated representative peers are unique; i.e. if different rules (e.g in different policies or different directions)
 // have same selectors, one representative peer is generated to represent both.
 // - note that :
-// - if the rule's namespaceSelector is nil, then the representative pod is created in the policy's Namespace (as it is a real namespace)
+// - this func generates only representative pods (does not generate representative namespaces);
+// each representative pod stores in its data its matching namespaceSelector and podSelector
+// - if the rule's namespaceSelector is nil and the policy is a namespace scoped (NetworkPolicy), then the representative pod is created in
+// the policy's Namespace (as it is a real namespace)
+// - if the rule's namespaceSelector is nil and the policy is cluster-scoped (ANP or BANP), then the representative pod will
+// store an empty namespaceSelector (matches all namespaces in the cluster) in its data
```
adisos, line 34: is this reflected in the output as "all pods with labels... in all namespaces? "

pkg/netpol/eval/exposure.go
```diff
@@ -26,13 +26,19 @@ import (
...
 // - if the rule's namespaceSelector is not nil, no representative namespace will be generated (representative pod has empty namespace name)
 // anyway, the representative pod will store the namespace data.
 func (pe *PolicyEngine) generateRepresentativePeers(selectors []k8s.SingleRuleSelectors, policyNs string) (err error) {
 	for i := range selectors {
 		podNs := "" // by default: representative peer has no namespace; (don't generate representative namespaces)
-		if selectors[i].NsSelector == nil {
+		// note that policyNs is "" (empty) for cluster-scoped policies (ANP and BANP) - means podNs kept empty in this case
+		if selectors[i].NsSelector == nil && policyNs != "" {
```
adisos, line 41: ` policyNs == ""` does not necessarily indicate that the policy is  cluster-scoped?
better rely on the policy type ?

adisos (commented):
initial comments/questions, did not review all files yet

pkg/netpol/connlist/exposure_map.go
```diff
@@ -123,14 +123,24 @@ func (ex *exposureMaps) addConnToExposureMap(pe *eval.PolicyEngine, allowedConne
 	if !ok { // should not get here
 		return errors.New(netpolerrors.ConversionToConnectionSetErr)
 	}
-	// check if the connection is contained in an entire cluster connection; if yes skip; if not store the connection data
-	contained, err := connectionContainedInEntireClusterConn(pe, peer, allowedConnSet, isIngress)
+	// check if the peer is exposed to entire-cluster and if connection equals the entire cluster connection;
+	equals, isExposed, err := connectionEqualsEntireClusterConn(pe, peer, allowedConnSet, isIngress)
```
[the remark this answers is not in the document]
shireenf-ibm: opposite, equals instead of contained-in (it was containedIn, where actually if it was not larger it was equals)
if there is a rule that allows any connection to entire-cluster when we compute allowed conns between peers, the result contains this connection ( so if it is same as entire-cluster connection, there is no need to add it as "separate" exposure).
if the connection is larger (i.e same conn to entire-cluster + other conn - it will be added as another exposure (was and yet))
the new case - with ANP we have deny rules that may be prior to the "allow to entire-cluster" :
so if a real-peer is exposed to entire-cluster on all-TCP
but there is a rule "deny to new-ns on TCP90" --- the exposure to "new-ns" is all TCP but TCP 90 which is different than the exposure to entire-cluster even it is containedIn it.

pkg/netpol/connlist/connlist.go
```diff
@@ -609,8 +609,8 @@ func (ca *ConnlistAnalyzer) getConnectionsBetweenPeers(pe *eval.PolicyEngine, sr
 					return nil, nil, err
 				}
 			}
-			// skip empty connections
-			if allowedConnections.IsEmpty() {
+			// skip empty connections unless one of the peers is representative
+			if allowedConnections.IsEmpty() && !(pe.IsRepresentativePeer(srcPeer) || pe.IsRepresentativePeer(dstPeer)) {
```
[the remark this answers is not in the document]
shireenf-ibm: i added it so i can handle the following case:
assume we have an ANP that denies all conns from a -> b ("b" is a representative peer)
and another (with lower priority) rule that allows all conns from a to entire-cluster

i.e "a" is exposed to entire-cluster on all conns except to b (conns are empty to b)
this was added so I can add the entry a->b in the exposure map (with "no conns" ) and display it in the output 
see example: "tests/exposure_test_with_anp_9"

pkg/netpol/connlist/exposure_map.go
shireenf-ibm: see example `tests/exposure_test_with_anp_9` and its document in `connlist_test.go`

pkg/netpol/eval/exposure.go
shireenf-ibm: yes (as previously)

pkg/netpol/eval/exposure.go
shireenf-ibm: for all current cases, it does (as the code is written ) and because an NP without namespace would raise error before getting here; we can also change to be more "accurate and sure"

pkg/netpol/eval/check.go
```diff
@@ -270,82 +270,163 @@ func (pe *PolicyEngine) allAllowedConnectionsBetweenPeers(srcPeer, dstPeer Peer)
...
 	}
-	// optimization: if all the conns between src and dst were determined by the ANPs : return the allowed conns
-	if anpCaptured && anpConns.DeterminesAllConns() {
-		return anpConns.AllowedConns, nil
+	// optimizations
+	podExposureUpdatedFlag := false
+	anpConnsDeterminedAllFlag := false
+	// if all cluster-wide conns were determined by the ANP then update the selected peer's cluster wide exposure
+	if pe.exposureAnalysisFlag && anpExposure.layerConns.DeterminesAllConns() {
+		podExposureUpdatedFlag = true
```
adisos, line 316: isn't exposure to external ip-blocks determined by regular netpols ?

pkg/netpol/eval/check.go
```diff
@@ -270,82 +270,163 @@ func (pe *PolicyEngine) allAllowedConnectionsBetweenPeers(srcPeer, dstPeer Peer)
...
+		updatePeerXgressClusterWideExposure(anpExposure.layerConns.AllowedConns, src, dst, isIngress)
+	}
+	// if all the conns between src and dst were determined by the ANPs : return the allowed conns
+	if anpConns.layerConns.DeterminesAllConns() {
+		if !pe.exposureAnalysisFlag || podExposureUpdatedFlag {
+			return anpConns.layerConns.AllowedConns, nil
+		}
+		// else : exposure-analysis is on and still not determined continue
+		anpConnsDeterminedAllFlag = true
```
adisos, line 325: how can get here?

pkg/netpol/connlist/exposure_map.go
shireenf-ibm: there is also an example with real-pod `tests/exposure_test_with_anp_10_with_real_pod` (when a real is exposed to entire-cluster, but the connection to another real-pod is blocked);
with real pods connlist-output gives as the indication 
(related also to this [comment](https://github.com/np-guard/netpol-analyzer/pull/486#discussion_r1929837225)

pkg/netpol/eval/check.go
shireenf-ibm: not only;
egress rules in (B)ANP can also determine conns to IPs (networks field)

pkg/netpol/eval/check.go
shireenf-ibm: because if a->b allows all (or at least determined all conns (ports) from ANPs) does not necessary mean that all exposure to entire cluster was also determined from ANPs .. so need to analyze also NPs / BANP for exposure

adisos: not sure I follow why

pkg/netpol/eval/check.go
adisos: can `podExposureUpdatedFlag` be set to true, if possibly more exposure should be inferred from regular netpols?

pkg/netpol/eval/check.go
shireenf-ibm: example:
ANP denies all conns from a -> b  (ANP determined all conns to b; if exposure analysis is off no need to continue to NP/BANP)
But:
NP exposes A to entire-cluster on all-conns (if we don't continue - this would not be computed and will have wrong results as exposure to entire-cluster is computed once for each direction)

pkg/netpol/eval/check.go
shireenf-ibm: i think this is related to the other [comment](https://github.com/np-guard/netpol-analyzer/pull/486#discussion_r1930113088) the answer is yes

adisos: it's not clear what do these flags mean?  where is it documented?

pkg/netpol/eval/check.go
adisos: ok. might be useful to add this example in comment as documentation, so it is clear.

pkg/netpol/eval/check.go
adisos: consider removing "entire-world" fields from exposure analysis structs, and mention in documentation that exposure analysis only computes exposure to entire-cluster (for external ip-blocks just copying the conn-list results).

pkg/netpol/eval/exposure.go
adisos: better rely on policy type, so the code is more robust

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
adisos: please add example/explanation in code

pkg/netpol/connlist/connlist.go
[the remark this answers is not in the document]
adisos: ok, please add comment in code explaining this

pkg/netpol/eval/check.go
shireenf-ibm: opened #488

pkg/netpol/connlist/connlist.go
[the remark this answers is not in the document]

adisos (commented):
for test `tests/exposure_test_with_anp_1/ ` , why isn't workload `b` mentioned in the output as "Workloads not protected by network policies:" ?

shireenf-ibm:
> for test `tests/exposure_test_with_anp_1/ ` , why isn't workload `b` mentioned in the output as "Workloads not protected by network policies:" ?

because the subject of the ANP is all `hello-world` pods
so it is protected and exposed to `slytherin` but since it is also exposed to `entire-cluster` on all connections (there is no deny rule/ any limiting policy selecting workload-b) then exposure to `slytherin` is actually contained in the exposure to entire-cluster

pkg/netpol/connlist/exposure_map.go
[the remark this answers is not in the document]
shireenf-ibm: done

pkg/netpol/eval/check.go
shireenf-ibm: done

adisos:
> > for test `tests/exposure_test_with_anp_1/ ` , why isn't workload `b` mentioned in the output as "Workloads not protected by network policies:" ?
> 
> because the subject of the ANP is all `hello-world` pods so it is protected and exposed to `slytherin` but since it is also exposed to `entire-cluster` on all connections (there is no deny rule/ any limiting policy selecting workload-b) then exposure to `slytherin` is actually contained in the exposure to entire-cluster

I see.. it's a bit misleading because its connectivity has the same effect as "not protected".. can open issue with this example to consider enhancements later.

pkg/netpol/eval/exposure.go

shireenf-ibm:
> exposure_test_with_anp_1/

done #489

Issue #455, opened 2024-12-01
[ANPs] admin-netpols tests coverage 


adisos:
when finalizing development of admin netpol support, revisit tests code coverage 
(conn-list, diff, eval, exposure analysis, expainability)

PR #494, opened 2025-01-28
Issue 455 admin netpols test coverage

shireenf-ibm:
#455 

* the only `err` paths that are not tested as bad-paths, are the errors returned from funcs related to `eval` command (i.e. bad-path tests for `eval`)

* commits are explained, each contains one "item"


pkg/netpol/internal/common/connectionset.go
```diff
@@ -171,26 +171,6 @@ func (conn *ConnectionSet) Contains(port, protocol string) bool {
 	return false
 }
 
-// ContainedIn returns true if current ConnectionSet is contained in the input ConnectionSet object
-func (conn *ConnectionSet) ContainedIn(other *ConnectionSet) bool {
```
adisos, line 175: better keep this function?

pkg/netpol/eval/resources.go
```diff
@@ -934,6 +934,10 @@ func (pe *PolicyEngine) addRepresentativePod(podNs string, objSelectors *k8s.Sin
 			// if the objSelectors.NsSelector is nil and the podNs is empty, means inferred from an
 			// (baseline)AdminNetworkPolicy rule with nil namespaceSelector, which means that all namespaces match the rule;
 			// so as the RepresentativeNsLabelSelector will assign an empty namespaceSelector (matches all namespaces)
+			// * keeping this else although currently it is unreached - since `v1alpha1 Pods *NamespacedPod` field uses
+			// `metav1.LabelSelector` (and not pointers); which means it can not be nil, either empty or not.(i.e. already empty)
+			// so anyway generating a representative peer with empty namespaceSelector (all namespaces ) for
+			// cluster-scoped policy rule without namespaceSelector
```
adisos, line 940: not sure I follow this comment and why it cannot be reached by tests

adisos (commented):
looks good!
added few small comments.

pkg/netpol/internal/common/connectionset.go
shireenf-ibm: reverted

pkg/netpol/eval/resources.go
shireenf-ibm: the `pods *(NamespacedPod)[https://pkg.go.dev/sigs.k8s.io/network-policy-api@v0.1.5/apis/v1alpha1#NamespacedPod]`  contains references of `metav1.LabelSelector` not pointers (unlike regular netpols rules fields).
i.e. a rule of ANP with `pods` without `namespaceSelector`, its actually stored by `apis/v1alphha1` as empty namespaceSelector originally (and not a `nil` namespaceSlector).
so the only valid case a `namespaceSelector` would be nil is if inferred from regular netpol which must have a non-empty namespace-name. 
(`namespaces` field is a pointer - but if was nil, then the ANP rule is invalid and an error would be raised; so this has no meaning in this case).

adisos (approved):

Issue #502, opened 2025-02-12
support `explain` flag with `exposure` flag


adisos:

shireenf-ibm:
and then in this PR, I will :

- [ ] allow running  with both `exposure` and `explain` flags (remove the error)
- [ ] avoid adding `No Connections` lines between a peer and ip-block in the exposure-results 
- [ ] add tests with both flags with expected results (currently only with txt output, didn't see support of explain with other output formats, if at all possible)

PR #503, opened 2025-02-12
`list`  support explain flag with exposure flag

shireenf-ibm:
#502 
- ensured that the exposure results section was not affected (and explanation contains only real peers and ip-blocks)


shireenf-ibm:
hi @adisos ,
I found out the the netpol's exposure to entire-cluster data is also updated with the implying rules
so in case of using the exposure optimization when the "exposure to entire cluster" allow all connections - the rule that indicates this is printed to the explain report (and same output for with and without `--exposure` is written)


--

so for the mean-while i have not turned off the optimization (when explain is on)
we'll see if Tanya has another case / issue with this, then I'll turn off it.

--
(tried again to run all of the `explanation_tests` : with `--explain` , `--exposure` and `--exposure --explain` >> all the outputs showed that running with both didn't affect the original results when running only one

adisos (approved):
looks good

Issue #505, opened 2025-02-13
[conn-list] additional flags to consider


adisos:
* `--focus-direction` {ingress/egress}" (no default) : filters the connections of the focus-workload on the  given direction.
`--focus-direction` has no meaning without `focusworkload`.
can add a warning if this flag is used without focusworkload, and ignore it.
example: 
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --focusworklod mybar --focus-direction egress ` (get only egress conns)
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --focusworklod mybar --focus-direction ingress ` (get only ingress conns)
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --focusworklod mybar` (get both ingress and egress conns) 

* `focusworkload-peer` : filters only  connections between `focusworkload` to `focusworkload-peer`
(a certain direction can be set with the previous flag)
example: 
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --focusworklod mybar  --focusworkload-peer myfoo`

* current `--focusworkload` flag, supporting comma-separated workloads (e.g. foo,bar)
example : `k8snetpolicy list --dirpath tests/anp_banp_blog_demo --focusworklod  mybar,mybaz`
(this applies also to `focusworkload-peer`)

*  `--explain-only`, that can have one of the values {"allow", "deny"} 
examples :  
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo  --explain --explain-only allow`  (explain only allowed conns)
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --explain --explain-only deny` (explain only denied conns)
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo --explain` (explain both allowed and denied conns)
(give a warning if this flag is used without `--explain` flag, and ignore it)

* `--focus-conn` : filter connectivity list (with/without explain) to the given connection specified.
value format  [protocol-port] (Example: `tcp-80` / `udp-53` )
examples:
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo  --explain --focusworklod mybar --focus-conn tcp-80`
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo  --focusworklod mybar --focus-conn tcp-80`
`k8snetpolicy list --dirpath tests/anp_banp_blog_demo  --focus-conn tcp-80`

flags summary:
- `--focus-direction` [ingress/egress]
[...]

PR #509, opened 2025-02-16
additional flags - focus-direction

shireenf-ibm:
#510

* first commit -support the new `--focus-direction` flag


adisos:
looks good.
let's change this flag so it has only `[ingress /egress]` values.
so that for running without specific direction, this flag should not be used.

pkg/cli/list.go
```diff
@@ -133,6 +138,8 @@ defined`,
 	// Use PersistentFlags() for flags inherited by subcommands or Flags() for local flags.
 	c.Flags().StringVarP(&focusWorkload, "focusworkload", "", "",
 		"Focus connections of specified workload in the output (<workload-name> or <workload-namespace/workload-name>)")
+	c.Flags().StringVarP(&focusDirection, "focus-direction", "", "",
+		"Filter connlist of specified workload for input direction {ingress, egress, both (default both)}")
```
adisos, line 142:
```suggestion
		"Focus connections of specified workload on one direction, applies only when focusworkload is used" {ingress/egress}
```

shireenf-ibm: done + change this flag so it has only [ingress /egress] values in code

pkg/internal/netpolerrors/netpol_errors.go
```diff
@@ -169,3 +169,7 @@ func ANPsWithSameNameErr(anpName string) string {
 func NPWithSameNameError(npName string) string {
 	return fmt.Sprintf("NetworkPolicy %q already exists. %s", npName, uniquenessRequest)
 }
+
+func FocusDirectionNotSupported(focusDirection string) string {
+	return focusDirection + " focus direction is not supported."
```
adisos, line 174: change to "invalid focus-direction value: <val> "

pkg/netpol/connlist/connlist.go
```diff
@@ -536,6 +559,10 @@ func (ca *ConnlistAnalyzer) getConnectionsList(pe *eval.PolicyEngine, ia *ingres
 
 	excludeIngressAnalysis := (ia == nil || ia.IsEmpty())
 
+	// validate focus-direction
+	if err := ValidateFocusDirectionValue(ca.focusDirection); err != nil {
```
adisos, line 563: is this the first place where this can be validated?

adisos:
> #505
> 
> * first commit -support the new `--focus-direction` flag

let's have separate PR for each sub-task in the issue
(can convert to sub-issues)

pkg/netpol/connlist/connlist.go
shireenf-ibm: no it is not,
if you run it from command-line, the validate would be done in the `PersistentPreRunE` of list command (see cli/list.go)
but i have added it here too, in case someone is using the API from a "script" for example;
this func `getConnectionsList` is the common for all the API funcs (i.e. `ConnlistFromDirPath`, `ConnlistFromResourceInfos` and `ConnlistFromK8sClusterWithPolicyAPI`)

pkg/internal/netpolerrors/netpol_errors.go
shireenf-ibm: done

shireenf-ibm:
I've realized that the cycle life of `VarP` is different than `StringVarP` 
so a `VarP` flag should be initialized or returned to its reset value before running the new (next) command

adisos:
looks good.
Few more small comments.

- let's be consistent in flags description:

```
--focus-direction string   Focus connections of specified workload on one direction, applies only when focusworkload is used {ingress/egress}
--output string            Required output format (txt,json,dot,csv,md) (default "txt")
```

should be changed to
```
--focus-direction string   Focus connections of specified workload on one direction, applies only when focusworkload is used; must be one of ingress,egress
--output string            Required output format; must be one of txt,json,dot,csv,md (default "txt")
```



- the cli error currently for invalid arg is:
`Error: invalid argument "ingressss" for "--focus-direction" flag: invalid focus-direction value: ingressss `

can be changed to:
`invalid argument "ingressss" for "--focus-direction" flag: must be one of ingress,egress`

pkg/netpol/connlist/connlist.go
adisos: so maybe move it to the beginning of this function?

adisos:
looks good,
can you also update the description of the `--output` flag as in my prev comment?

README.md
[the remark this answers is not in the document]
shireenf-ibm: done + output desc.

adisos (approved):

Issue #480, opened 2025-01-16
[conn-list] Generate connlist if labels gap between pods with same owner does not affect the result


shireenf-ibm:
currently, if there are pods with same owner in the manifests but with different labels set;
an error is raised and the the analyzer does not produce a connlist output.
In this issue, if the gap between the pods' labels does not affect the results;
i.e. if there are no policies or if the existing policies don't contain selectors with labels from that gap then the analyzer should proceed and generate a connlist output (such pods can be associated to the same owner workload for analysis).

PR #552, opened 2025-03-18
Issue 480: cases to enable connectivity analysis even when there is a labels-gap between pods of same owner-reference

shireenf-ibm:
#480 

currently:
- code with first approach : generating error if its not possible to analyze-connectivity when `createPodOwnersMap` (by looping policies' selectors)

- multiple examples

~- [ ] possible to enhance the performance by generating the errors on the flow of computing allowed-conns between peers~
moved to a new issue #553

### DIFF
--- a/pkg/netpol/connlist/connlist_test.go
+++ b/pkg/netpol/connlist/connlist_test.go
@@ -2003,6 +2003,24 @@ var goodPathTests = []struct {
 		outputFormats:          []string{output.DefaultFormat},
 		supportedOnLiveCluster: true,
 	},
+	{
+		// in following example we have 2 pods which belong to same owner;
+		// the pods have different labels;
+		// however, since there are no policies in the input resources
+		// a connectivity report is generated (the labels diff does not affect the analysis)
+		testDirName:            "example_pods_w_same_owner_and_labels_gap_no_policy",
+		outputFormats:          []string{output.DefaultFormat},
+		supportedOnLiveCluster: true,
+	},
+	{
+		// in following example we have 2 pods which belong to same owner; the pods have different labels;
+		// however, since the policy rule selector (which selects those pods) uses a label-selector which is same in both pods
+		// i.e. not from the gap:
+		// a connectivity report is generated (the labels diff does not affect the analysis)
+		testDirName:            "example_pods_w_same_owner_and_labels_gap_with_policy_selector_not_in_gap",
+		outputFormats:          []string{output.DefaultFormat},
+		supportedOnLiveCluster: true,
+	},
 }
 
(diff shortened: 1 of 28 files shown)


adisos:
new error message is not so readable:
```
./bin/k8snetpolicy  list --dirpath tests/example_pods_w_same_owner_and_labels_gap_with_bad_match_expression/
Error: Input resources are not supported for connectivity analysis. Found Pods of the same owner default/internal-security but with different set of labels. And found a policy containing selectors from the different labels set. Pod default/pod-2 has label app=query-selector, and Pod default/pod-1 has label app=command-selector. While NetworkPolicy: default/use_common_labels_same_values contains selector &LabelSelector{MatchLabels:map[string]string{},MatchExpressions:[]LabelSelectorRequirement{LabelSelectorRequirement{Key:app,Operator:In,Values:[query-selector],},},}
```

Consider rephrasing to something like the following:
```
Found Pods of the same owner workload `default/internal-security` and with differences in their labels, while a network policy contains selectors impacted by this gap. 
Pod `default/pod-2 has` label app=query-selector, and Pod `default/pod-1` has label app=command-selector, while NetworkPolicy: `default/use_common_labels_same_values` contains selector `{Key:app,Operator:In,Values:[query-selector]}`.
```

shireenf-ibm:
> new error message is not so readable:
> 
> ```
> ./bin/k8snetpolicy  list --dirpath tests/example_pods_w_same_owner_and_labels_gap_with_bad_match_expression/
> Error: Input resources are not supported for connectivity analysis. Found Pods of the same owner default/internal-security but with different set of labels. And found a policy containing selectors from the different labels set. Pod default/pod-2 has label app=query-selector, and Pod default/pod-1 has label app=command-selector. While NetworkPolicy: default/use_common_labels_same_values contains selector &LabelSelector{MatchLabels:map[string]string{},MatchExpressions:[]LabelSelectorRequirement{LabelSelectorRequirement{Key:app,Operator:In,Values:[query-selector],},},}
> ```
> 
> Consider rephrasing to something like the following:
> 
> ```
> Found Pods of the same owner workload `default/internal-security` and with differences in their labels, while a network policy contains selectors impacted by this gap. 
> Pod `default/pod-2 has` label app=query-selector, and Pod `default/pod-1` has label app=command-selector, while NetworkPolicy: `default/use_common_labels_same_values` contains selector `{Key:app,Operator:In,Values:[query-selector]}`.
> ```

done

adisos:
>  possible to enhance the performance...

please open a separate issue for this, better consider address this separately and not in this PR.

adisos (approved):

PR #554, opened 2025-03-19
appending connlist-analyzer warnings to ca.errors to be also used by its API Errors func.

shireenf-ibm:
this PR is needed so we can see all warnings raised by the connlist-analyzer when used as an API / imported pkg


shireenf-ibm:
tested locally : now warnings on unsupported format for `explain` appear when using the API funcs

adisos (approved):

Issue #561, opened 2025-04-01
initial support of UserDefinedNetwork (UDN) objects


shireenf-ibm:
add support of UserDefinedNetwork (UDN) objects (as `Primary` Network)
only one primary UDN may be assigned to a namespace
 
general implementation-plan: 
* a namespace stores reference to its primary UDN; so its pods are isolated from pods in other UDNs/namespaces

useful links:
1. https://ovn-kubernetes.io/api-reference/userdefinednetwork-api-spec/#userdefinednetwork
2. https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/networking/multiple-networks#about-user-defined-networks
3. https://pkg.go.dev/github.com/ovn-org/ovn-kubernetes/go-controller/pkg/crd/userdefinednetwork/v1#UserDefinedNetwork

tasks that can be split to separate PRs:

- [x] support UDN + output updates #565 

- [x] add explainability when connection between peers is blocked since at least one is in an isolated udn + add tests + (update output labels if required) #568 

~- [ ] add tests with exposure-analysis : peer in a udn should not be exposed to any namespace (+update output labels if required)~ moved to #573

- [x] support ingress-analyzer and add test with ingress-controller  #570 
- [x] add tests with warnings (ignoring UDN)  #565 
- [x] add bad-path tests (returning errors) #572 

~- [ ]  update diff conns output with "[udn]" labels (a udn may be unchanged/added/removed)~ moved to #573

PR #568, opened 2025-04-16
Add explainability with udn

shireenf-ibm:
#561 

> * add explainability when connection between peers is blocked since at least one is in an isolated udn + add tests + (update output labels if required)


shireenf-ibm:
- [ ] group connections by UDN (similar to regular connlist)

- [ ] add section for denied connections because of UDN assertions (when at least one of the peers is isolated in a udn)

test_outputs/connlist/udn_test_1_tcp-90_explain_output.txt
```diff
@@ -0,0 +1,52 @@
...
+	Allowed TCP:[90] due to the system default (Allow all)
+
+----------------------------------------------------------------------------------------------------------------------------------------------------------------
+Connections between blue[udn]/webserver[Pod] => 0.0.0.0-255.255.255.255[External]:
+
+Allowed connections:
+	Allowed TCP:[90] due to the system default (Allow all)
+
+----------------------------------------------------------------------------------------------------------------------------------------------------------------
+Connections between blue[udn]/webserver[Pod] => green[udn]/webserver[Pod]:
```
adisos, line 18: why is this pair (mixing two udns) under `blue[udn]` header?

shireenf-ibm: it was a random decision to put it under the "src's udn"; in the mean while a connection which mixes two udns may appear in the report only in case we are using a "focus-*" flag to run the command 
(there are no allowed conns between peers in different networks, and denied cross network connection will not appear in the report unless using a specific focu-conn/workload flag)

adisos (approved):

PR #574, opened 2025-04-21
Bug fix for issue571

tanyaveksler:
Bug fix: when collecting multiple ANP connections, don't propagate explainability of `PassConns` into `AllowedConns` and `DeniedConns`.
Added netpol-demo to explainability test base.


adisos (approved):
looks good!

Issue #603, opened 2025-05-22
Support Secondary networks for pods


shireenf-ibm:
some useful links:

- https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/networking/multiple-networks#understanding-multiple-networks
- https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/virtualization/networking#secondary-network-config
- https://kubeops.net/blog/exploring-multus-an-advanced-networking-solution-for-kubernetes
- https://pkg.go.dev/github.com/k8snetworkplumbingwg/network-attachment-definition-client/pkg/apis/k8s.cni.cncf.io/v1#NetworkAttachmentDefinition

Issue #573, opened 2025-04-20
[UDN, NAD] future enhancements


shireenf-ibm:
possible enhancements for UDN to be considered:
- [x] [NAD, Multi-NP]: support selecting internal peers (pods/Vms ..) by their IP-addresses with `IPBlock` selector in a multi-network-policy's rule 
       Note that: currently secondary networks describe connections between workloads in the cluster-only
        - conns between external-ipblock-peers and internal-pods are attached to a primary network or the pod-network (depending on the primary network that the internal peer belongs to)
        - see example with VM : [here](https://kubevirt.io/2023/OVN-kubernetes-secondary-networks-policies.html)
        - and with pods : [here](https://github.com/openshift/multus-networkpolicy/blob/278ec20e795c3a590500e789716be7fcc4d7107b/e2e/tests/ipblock.yml)
#633 
        
- [x] 2 peers may connect with more than one network (for example : multiple secondary interfaces (if belong to them))
see example [here](https://github.com/openshift/multus-networkpolicy/blob/278ec20e795c3a590500e789716be7fcc4d7107b/e2e/tests/bond-cni.yml) (#622)

- [ ] update README with new features & prepare demo-presentation

PR #612, opened 2025-06-25
NAD and MultiNetworkPolicy support

shireenf-ibm:
#603 
### **PR Description :**

1. **Dependency Updates** (go.mod and go.sum)
- New Modules Added:
   * ` github.com/k8snetworkplumbingwg/multi-networkpolicy`
   * `github.com/k8snetworkplumbingwg/network-attachment-definition-client`
- Purpose: These additions enable support for `MultiNetworkPolicy `and `NetworkAttachmentDefinition (NAD) ` resources, which are essential for managing secondary networks in Kubernetes.

2. **Core Feature Enhancements**
- Support for New Kubernetes Resources; 
   - New Kinds Added:
      *  `MultiNetworkPolicy`
      *  `NetworkAttachmentDefinition`

3. **Network Policy Evaluation Enhancements**

-  Primary vs. Secondary Network Awareness
    -  Primary Networks: `UDN (User Defined Network), CUDN (Cluster UDN)`
    -  Secondary Networks: `NADs`
    -  New Logic:
        *  Determine if a pod belongs to a primary or secondary network.
        *  Identify isolation between pods based on network membership.
        *  Evaluate connectivity using either:
             - Traditional Kubernetes `NetworkPolicies` (for primary networks and pod-network)
             - `MultiNetworkPolicies` (for secondary networks)
 
- New Functions Introduced:
   - `getPeerPrimaryNetwork` 
   - `getPeerSecondaryNetworks`
   - `findCommonSecondaryNetworkForPeersPair` 
   - `allAllowedXgressConnectionsByMultiNetpolsCRDs`

4. **Testing and Validation**
- New Test Cases:
   - Added multiple `nad_test_*` scenarios to validate connectivity across `NADs` and `MultiNetworkPolicies`.
   - Focused tests on ingress/egress rules, namespace selectors, and protocol/port combinations.
   - Some tests involve `UDN` and `NAD`

5. **Output and Visualization Improvements**
- Enhancements:
       - Group peers by `NAD`s in visual outputs.
       - Annotate connections with network context (e.g., pod network, `UDN`, `NAD`).
       - Include `networkData` in connection metadata for better traceability.

6. **Code Refactoring**
- Removed:
   - `primaryUdnNamespaces` field from `ConnlistAnalyzer` (now inferred dynamically).
- Modularized:
   - Shared logic for port parsing, selector matching, and explanation generation moved to `netpols_common.go`.

7. **Explanation and Debugging**
- Improved Explainability:
   - Clear reasons for allowed/denied connections.
   - Differentiation between default rules, explicit policies, and network isolation.

**Key Concepts**
- Primary network takes precedence: If a pod is in a UDN or CUDN, that becomes its default network.
- Secondary networks are additive: Pods can communicate over NADs if both are attached to the same NAD and a MultiNetworkPolicy allows it (and both are not isolated by any primary network)
- for this version, in the output report : allowed conns of pair of two peers `src` and `dst`  may appear in one network only;
since the `pe.AllAllowedConnectionsBetweenWorkloadPeers(srcPeer, dstPeer)` returns only one `connectionSet` (a connection in one network).
  * (if the peers belong to more than one secondary network, only one is reported for now)
  * ( if the peers are in a UDN and also belong to a secondary network, the primary UDN connection only appears; should both appear in the follow up issue?)
  * ( if one is isolated by a UDN, and both belong to a secondary network, no conns allowed between the peers; see first key concept above)
  *  if both src and dst are not in another primary/ secondary network, then the conns between them are in the pod-network

--------------------------------------------------
**Todo in a follow-up issues:**
- a TODO task was added to #573 

> [NAD, Multi-NP]: support selecting internal peers (pods/Vms ..) by their IP-addresses with IPBlock selector in a multi-network-policy's rule
Note that: currently secondary networks describe connections between workloads in the cluster-only.
conns between external-ipblock-peers and internal-pods are attached to a primary network or the pod-network (depending on the primary network that the internal peer belongs to)

- TODO : behavior when more than one secondary network is common for pods (added to #573 too)
Note that : 
currently when computing allowed conns between peers, a single connection is returned (connection in one network)
in this follow-up issue, two peers may connect in multiple networks, each with a specified connection;

- TODO: bad path check (NADs with same name must have same network config) + bad path tests (added to #573 too)


adisos (approved):

adisos:
one comment about output (can be considered in a separate issue/PR ) : 
- for various tests, sometimes default pod network appears in the output and sometimes not (e.g.  tests/udn_and_vms_test_1 , nad_test_1/ ,  tests/nad_test_9/ ) . I think it should always appear (even if no connection is allowed)  - if the multi-network option is  on. 
if it has empty connection lines, can add in the output why (either no input workload has pod-network as primary network, or all connections are not allowed for the workloads in this network).

shireenf-ibm:
`golint` failing issue is resolved in #616 , so it will be fixed automatically after merging #616 with current branch

adisos (dismissed):

Issue #634, opened 2025-09-02
run netpol-analyzer with debug mode


shireenf-ibm:
- [x] **Resource Utilization Logging**:  #648
    - Emit detailed debug messages indicating which resources were involved in the analysis.
    - This will provide visibility into the scope and selection of resources, aiding in validation and troubleshooting.

- [ ] **Gap Analysis Logging**
     - After resource participation is established, log debug messages that highlight discrepancies between:
             - The actual resources detected or available
             - The objects explicitly inserted in the analysis
     - This will help identify missing, mismatched, or unaccounted-for elements, improving the accuracy and completeness of the analysis.

PR #648, opened 2025-10-01
first debugging task

shireenf-ibm:
#634 - part 1


adisos (commented):
can you also print the yaml manifest of the processed resource that is added (not the original input manifest)?

shireenf-ibm:
> can you also print the yaml manifest of the processed resource that is added (not the original input manifest)?

since writing the whole processed yaml will be so long and contains multiple not relevant fields; i chose to print the Spec's yaml of the objects.
the only exception is for workload objects (we handle them as interface{} objects so can't directly get their spec); i print the workload's labels.

- tested locally by:  changing an indentation of one's network-policy labels (in the input np yaml)
the debugger printed `matchLabels: {}` (ignored the labels with the wrong indentation) - as expected

adisos:
> > can you also print the yaml manifest of the processed resource that is added (not the original input manifest)?
> 
> since writing the whole processed yaml will be so long and contains multiple not relevant fields; i chose to print the Spec's yaml of the objects. the only exception is for workload objects (we handle them as interface{} objects so can't directly get their spec); i print the workload's labels.
> 
> * tested locally by:  changing an indentation of one's network-policy labels (in the input np yaml)
>   the debugger printed `matchLabels: {}` (ignored the labels with the wrong indentation) - as expected

ok, how about also an additional flag to dump all processed  yaml resources that are used for the analysis, into a yaml output file?

shireenf-ibm:
> > > can you also print the yaml manifest of the processed resource that is added (not the original input manifest)?
> > 
> > 
> > since writing the whole processed yaml will be so long and contains multiple not relevant fields; i chose to print the Spec's yaml of the objects. the only exception is for workload objects (we handle them as interface{} objects so can't directly get their spec); i print the workload's labels.
> > 
> > * tested locally by:  changing an indentation of one's network-policy labels (in the input np yaml)
> >   the debugger printed `matchLabels: {}` (ignored the labels with the wrong indentation) - as expected
> 
> ok, how about also an additional flag to dump all processed yaml resources that are used for the analysis, into a yaml output file?

added a new flag that may be used in one of the following:

1. `--dump-processed-yaml`: dumps processed yaml resources to `./processed-resources.yaml`.
2. `--dump-processed-yaml=<file-path>`: allows specifying a custom file path for the dump (the `=` is required)

